PR #16, opened 2018-06-27
fix-deprecation-prompt

natecavanaugh:
/cc @jbalsas @izaera


jbalsas:
Hey @natecavanaugh, not sure what's wrong with the CI now... 🤔 

I'm merging so we can go ahead and keep adding stuff.

Thanks!

PR #20, opened 2018-07-10
Handles inexistent npm module exception properly

jbalsas:
Hey @natecavanaugh, this should fix the issues when trying to locate themes or themelets from npm registry. As far as I can tell, some dependencies needed to be updated and some APIs changed.

All tests are passing locally, we can use this to try to figure out the CI issues as well.

Finally, as yoy can see, I'm sending this to the `develop` branch. This is the git flow we're following here:
- `master` contains the code of the latest published version on npm
- `develop` contains the current work in progress
- We use a merge commit to merge into `develop`.
- When we want to release, we send a PR to `master` from `develop` and if all tests pass, then we merge and release from `master`

### DIFF
--- a/packages/generator-liferay-theme/package-lock.json
+++ b/packages/generator-liferay-theme/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "generator-liferay-theme",
-  "version": "8.0.0-alpha.6",
+  "version": "8.0.0-beta.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -9,8 +9,8 @@
       "resolved": "https://registry.npmjs.org/CSSselect/-/CSSselect-0.4.1.tgz",
       "integrity": "sha1-+Kt+H4QYzmPNput713ioXX7EkrI=",
       "requires": {
-        "CSSwhat": "0.4",
-        "domutils": "1.4"
+        "CSSwhat": "0.4.7",
+        "domutils": "1.4.3"
       }
     },
     "CSSwhat": {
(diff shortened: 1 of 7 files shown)


natecavanaugh:
@jbalsas Thanks for the heads up on the git flow process. We used it on the old Lexicon repo, but since it's the default, it will let me update locally so I make sure I'm on the right version :)

About the CI changes, I think things should be working (at least the latest build went through after some updates to eliminate the failing test and prettier formatting.

Re: the changes here, they look good to me, though just one quick question:
Since this uses features that will break in the maintenance LTS of node (async/await), should be specify a minimum node version in package.json? Mainly, I'd rather people get the hint that the version of node required needs to be higher when they go to install rather than it installing and reporting errors later on.
Just a thought though :)

Everything else, LGTM!

jbalsas:
Hey @natecavanaugh, that's a good point!

We ran the frontend training last week, and it turns out that this already won't run in node<6, I think, so it's definitely worth updating it.

Can you think of a way to use it so we can limit how high a node version we might need? I tried simple promises but didn't seem to work. 🤔

natecavanaugh:
@jbalsas Wierd... I was under the impression that await (implicitly) relied on Promises already.
Can you go into detail as to why it didn't work with just plain promises?

Either way, I think node < 6 shouldn't even be a concern (though I think the package.json should at least error it quickly).

Thanks again for all of this help!

jbalsas:
Nothing, really, I guess I was just tired... I tried promises, but used callback style instead of promise catch 😂 

I've updated it and added `node>=8.0.0` to the engines property. I think we might be able to relax it to `6.0`, but based on how long it's going to stay around and being `8.x.x` the current active LTS version, I think it should be fine.

jbalsas:
Hey @natecavanaugh, CI looks good, so I'm going to shamelessly auto-merge since you okey'd it before... but please, review it if you have a chance so we can fix anything that I might've done wrong here :)

Thanks for the help!

PR #32, opened 2018-07-24
Improve information regarding theme/themelet npm search

izaera:


packages/liferay-theme-tasks/lib/look_and_feel_util.js
```diff
@@ -283,7 +279,7 @@ function mergeThemeElementByValue(themeElements, baseThemeElements) {
 }
 
 // Export private methods when in tests
-if (jest) {
+if (typeof jest !== 'undefined') {
```
jbalsas: Why don't we simply export it always using the regular ES6 export mechanism?

packages/liferay-theme-tasks/lib/theme_finder.js
```diff
@@ -139,23 +171,35 @@ function isLiferayThemeModule(pkg, themelet) {
...
-			liferayTheme &&
-			(themelet ? liferayTheme.themelet : !liferayTheme.themelet);
+		return isLiferayThemeModule.OK;
 	}
 
-	return retVal;
+	return isLiferayThemeModule.NO_PACKAGE_JSON;
 }
 
+Object.assign(isLiferayThemeModule, {
```
jbalsas: Why don't we simply create a different object for constants and export it? merging it with a function and then using it through there is a bit confusing 🤔

packages/liferay-theme-tasks/lib/theme_finder.js
```diff
@@ -215,29 +250,29 @@ function searchGlobalModules(config, cb) {
...
 
-			cb(themeResults);
+			cb(moduleResults);
 		});
 	});
 }
 
 // Export private methods when in tests
-if (jest) {
+if (typeof jest !== 'undefined') {
```
jbalsas: Same here, why don't we export everything properly? Is this really meant to be private? If it is, should we be testing it like that? :D

packages/liferay-theme-tasks/lib/util.js
```diff
@@ -155,7 +155,7 @@ function validateCustomDependencyPath(customPath) {
 }
 
 // Export private methods when in tests
-if (jest) {
+if (typeof jest !== 'undefined') {
```
jbalsas: :)

packages/liferay-theme-tasks/lib/look_and_feel_util.js
izaera: We can export it always (I preferred to control visibility in a safer way but, hey!, this is javascript, an `_` is enough :-D ).

But... I don't want to use ES6 export because that would force us to transpile and this project is not transpiling yet. 

I don't have problems with babelifying another project, but if it is not yet needed, why doing it now?

jbalsas: 👍

packages/liferay-theme-tasks/lib/theme_finder.js
izaera: OK. That was the other option I thought of and I wasn't too sure on which one to use...

packages/liferay-theme-tasks/lib/theme_finder.js
jbalsas: We'll keep it as is for now based on the previous comment

packages/liferay-theme-tasks/lib/util.js
jbalsas: We'll keep it as is for now based on the previous comment

packages/liferay-theme-tasks/lib/theme_finder.js
izaera: Yes, it is meant to be private. Regarding if we should test it, I can argue for and against testing private methods... In this case the tests were already written, so I preferred to leave them as they were.

jbalsas: 👍

PR #40, opened 2018-07-27
Update links to documentation and stop using deprecated Google URL shortener

izaera:


packages/generator-liferay-theme/lib/divert/7.0/app_helpers.js
```diff
@@ -1,39 +0,0 @@
-var chalk = require('chalk');
```
jbalsas: Why don't we need these 2 anymore?

izaera: Because they contain the same code and have been moved to common. Common code was different but because it was a leftover  from 6.2

PR #41, opened 2018-07-27
Don't overwrite rubySass setting when upgrading theme

izaera:
I have sent this in a different PR because it seemed that the `rubySass` overwrite was made on purpose, so I'm not sure if this will break some other thing.

If unsure, just don't merge this PR but only [this one](https://github.com/liferay/liferay-themes-sdk/pull/40)


jbalsas:
Hey @natecavanaugh, do you remember if this was indeed intentional? It seems indeed a bit intrusive and counter-intuitive, but seems to signal that we wanted to remove `rubySass` usage.

Still, this might be necessary in Windows environments... 🤔 

I'm going to be merging, for now, we can always revert back if necessary if we figure out this as an error.

Thanks!

yuchi:
> Still, this might be necessary in Windows environments...

AFAIK libsass support for Windows should be considered stable now. @GianBe could you confirm this?

PR #74, opened 2018-08-31
Fixes #67 - Updates liferay-plugin-node-tasks to v2.0.0

### LINKED ISSUE #67: Generator-Liferay-Theme: Do not insist on app server directory
During the liferay-theme generator process the user has to enter the path to the app server directory. The assistant would not continue if the path was not valid.

We are running Liferay from a Docker container and not a local Tomcat instance and therefore there is no app server directory.

Please allow the assistant to finish if there is no such directory. Of course the default live reload tasks are not going to work.
Maybe add https://github.com/GlassOfWhiskey/docker-gulp-tasks-4-liferay to core.

jbalsas:


jbalsas:
Merging, NPM global registry search API seems to be down 😢

PR #87, opened 2018-09-12
Fixes #46 - Documents postcss option

### LINKED ISSUE #46: Add out-of-the-box support for PostCSS
From [use autoprefixer for css and sass](https://github.com/liferay/liferay-theme-tasks/issues/81) by @marcelmokos and [Apply postcss during build of a Liferay 7 theme](https://github.com/liferay/liferay-theme-tasks/issues/70) by @caneta

References on current approach:
- [styled-svg-sprite-theme](https://github.com/marcoscv-work/styled-svg-sprite-theme)
- [styled-postcss-theme](https://github.com/marcoscv-work/styled-postcss-theme)

jbalsas:
Hey @yuchi, @sbaechler, this is an initial attempt to add PostCSS by default to the theme compilation process.

Basically, it allows you to add the following to your `package.json`:

```json
"liferayTheme": {
	[...]
	"postcss": [
		"autoprefixer",
		"postcss-modules"
	],
	[...]
}
```

Based on how common `autoprefixer` is, I thought we could bundle it with the tasks, so all that's needed to add this is the configuration. For other plugins, one should do `npm i postcss-modules` in their theme project.

Based on my experimentation, the range of plugins that can be used might be a bit limited, because we need to get through the SASS compilation first, and it will complain about things it doesn't understand (special syntax and so on...) so by the time the plugins get in, it will be too late.

I've tried the different SASS parsers for PostCSS and none seem to do the trick with our code, so I think this is the best we can do at the moment.

This also seems to fix https://github.com/liferay/liferay-themes-sdk/issues/83, or at least I'm seeing the sourcemaps (which we're [planning on enabling by default](https://github.com/liferay/liferay-themes-sdk/issues/86)) being properly generated.

What do you think?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,129 @@
+# Contributing Guidelines
+
+If you wish to contribute to Liferay Themes SDK these guidelines will be
+important for you. They cover instructions for setup, information on how the
+repository is organized, as well as contribution requirements.
+
+## Setup
+
+TBD
+
+## Repo organization
+
+TBD
+
+## Pull requests & Github issues
+
+* All pull requests should be sent to the `develop` branch, as the `master`
+branch should always reflect the most recent release.
+* Any merged changes will remain in the `develop` branch until the next
+scheduled release.
+* The only exception to this rule is for emergency hot fixes, in which case the
+pull request can be sent to the `master` branch.
+* A Github issue should also be created for any bug fix or feature, this helps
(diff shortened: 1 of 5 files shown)


jbalsas:
I'm going to be merging this tentatively for now. We can always tweak it a little bit if you guys have further feedback.

Thanks!

yuchi:
Missed the call for review. I’ll have a look as soon as I can.
/cc @crcarlo

PR #96, opened 2018-09-26
Fixes #72: Support Docker Development

### LINKED ISSUE #72: Support Docker Development
Based on https://github.com/liferay/liferay-themes-sdk/issues/67 by @sbaechler:

> We are running Liferay from a Docker container and not a local Tomcat instance and therefore there is no app server directory.
>
> Maybe add https://github.com/GlassOfWhiskey/docker-gulp-tasks-4-liferay to core.

@sbaechler, we'd be more than happy to take contributions if you're up for it!

In case you can't contribute, we'll definitely look at your work and try to add support for this soon.

GlassOfWhiskey:
## Added Docker support for `deploy` and `watch` tasks

Some Docker specific tasks have been created inside `tasks/deploy.js` and `tasks/watch.js` to transfer files and commands to a Docker container. Standard task pipelines have been modified accordingly, based on the concept of `deploymentStrategy` defined by the developer during the `init` task.
The new features have been tested with `glassofwhiskey/liferay-portal:7.1-ce-ga1-dev` container. Please note that `watch` task can be supported only if remote GoGoShell access has been enabled inside the target container.

## Added `dockerThemesDir` option

Determines where temporary files, necessary to launch the `watch` task, are created. Liferay user must have read and write permissions on the selected directory. The default value is `/home/liferay/tmp/themes`.

### dockerCopy

Copies a folder or a set of files from the host fs to the container fs. Since `docker cp` command can only create files owned by root inside Docker Container, a more sophisticated way to transfer files has been adopted.

### dockerExec

Executes a command inside Docker Container and returns the child process to extract output from stdio or errors from stderr.

### getPath

If the selected `deploymentStrategy` requires a deployment on Docker container, path utilities will refer to `posix` rules, independently of the host OS. Note that this assumes that only Unix Containers will be used. If support must be extended to Windows containers, this method must be modified accordingly.

## Set `notify: false` in BrowserSync options

BrowserSync notifications system seems not to be compatible with
SennaJS Screens feature.

BrowserSync tries to delete the notification div after a short delay,
but it searches such a div as direct child of `<body>`.
With SennaJS Screens, since Body element is a surface element,
the `#__bs_notify__` div element is moved inside the `surface1-default`
child, but this causes an exception to be thrown while removing the
element. Since it happens often but not always, the issue is probably
due to a race condition between the initialization of the two libraries.

As a consequence, the BrowserSync autoreload capability is compromised,
and the gulp watch task doesn't work as expected. Disabling BrowserSync
notify feature actually solves the problem.


jbalsas:
Just checked the tests, just some formatting failures, I'll push fixes later as soon as I test this. Hope to get this done tomorrow! 🙏

izaera:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

GlassOfWhiskey:
Suerte XD

Il giorno Lun 26 Nov 2018, 15:46 Ivan Zaera <notifications@github.com> ha
scritto:

izaera:
Hey @GlassOfWhiskey:

I've been playing around with your PR and docker and I've found some things that don't work in my environment and some others that I don't have quite clear. BTW, my environment is a Mac using `Docker version 18.09.0, build 4d60db4` and the official image `liferay/portal:7.1.0-ga1`.

IMO, any Docker feature added to this repo should work out of the box with Liferay's official images with no modications (if possible).

Regarding the deploy task, it doesn't work for me as `tar` is complaining with `short read` which, apparently means that the tar file is corrupted. I have changed the way to copy files and this seems to work for me:

```javascript
function dockerCopy(containerName, sourceFolder, destFolder, sourceFiles, cb) {
	if (_.isFunction(sourceFiles)) {
		cb = sourceFiles;
		sourceFiles = undefined;
	}

	if (!cb) {
		cb = function() {};
	}

	let packConfig = {
		dmode: parseInt(755, 8),
		fmode: parseInt(644, 8),
	};

	if (sourceFiles) {
		_.assign(packConfig, {
			entries: sourceFiles,
		});
	}

	tar.pack(sourceFolder, packConfig).pipe(
		es.wait(function(err, body) {
			if (err) throw err;

			const sourceTarFile = path.join(sourceFolder, 'docker-tar.tar');
			const destTarFile = '/opt/liferay/docker-tar.tar';

			fs.writeFileSync(sourceTarFile, body);

			let proc = childProcess.spawnSync('docker', [
				'cp',
				sourceTarFile,
				containerName + ':' + destTarFile,
			]);

			if (proc.error) {
				return cb(proc.error);
			}

			proc = childProcess.spawnSync('docker', [
				'exec',
				containerName,
				'/bin/tar',
				'x',
				'-f',
				destTarFile,
				'-C',
				destFolder,
			]);

			if (proc.error) {
				return cb(proc.error);
			}

			fs.unlinkSync(sourceTarFile);

			childProcess.spawnSync('docker', [
				'exec',
				containerName,
				'/bin/rm',
				destTarFile,
			]);

			cb();
		})
	);
}
```

I basically create the tar file locally, copy it to docker and extract it there to avoid issues with piping. I don't know why that works and piping the tar file as a stream to docker doesn't but that's what happens. It even happens running docker container's tar command outside node from a terminal.

Next thing is the watch task. I'll add another comment to the PR on that later. Stay tuned ;-)

izaera:
OK, let's go with watch. 

For the record, I have started my docker instance with the following command:

```sh
docker run -it -p 11311:11311 -p 8080:8080 --env LIFERAY_MODULE_PERIOD_FRAMEWORK_PERIOD_PROPERTIES_PERIOD_OSGI_PERIOD_CONSOLE="0.0.0.0:11311" <image id>
```

That makes gogo console start and listen to port 11311 which is then exported to docker's host. It also makes everything work without any need to map folders or modify the official images.

I have managed to make watch work once but it then fails, probably because of current bugs in the gogo console (7.1.1) and the wab handler (7.1.0) that prevent watch from working correctly. However, it seems to do its job.

Nevertheless, I would propose removing the `dockerThemesDir` option and hardcoding it to `/tmp`. I don't see the need to make temporary dirs configurable as they only serve internal purposes. Does it really make any sense to overwrite that folder?

The rest looks OK :-)

izaera:
To finish with: do we really need the change in .eslintrc https://github.com/liferay/liferay-themes-sdk/pull/96/commits/f5e4a3c86e2c8a266b754753e2c25d2cd120cb3c#diff-1dc6ee56b778cd91e0327b52aaeaa1b9 ?

izaera:
@GlassOfWhiskey how do you want to proceed? Do you want to make the changes that I proposed to your PR and test them in your env?

1. the copy fix 
2. the removal of `dockerThemesDir` if proved possible
3. restoring .eslintrc 

I can then redownload the PR and make sure it works in my env too...

Or if you prefer I can add my modification on top of your PR and resend...

GlassOfWhiskey:
Hola @izaera,
I will comment your three observations below. I will start from 2 and 3, because they are easier. The first is much more technical, so I will reserve a dedicate comment for it.

2. **The removal of `dockerThemesDir` option is doable**. The important thing is that the `liferay` user must have rw permissions on the involved folders tree, in order to avoid _Permission denied_ errors. I think that simply put themes inside `/tmp` folder works almost always (it is very unlikely that a developer absolutely needs to create folders with the same name inside `/tmp` for his/her project).

3. **The rollback of `.eslintrc` is doable**. I modified `.eslintrc` file because Windows _CRLF_ style gave me a lot of warnings. Then I discovered a way to explicitly set _LF_ new line characters inside Visual Studio Code, but I forgot to rollback the `.eslintrc`. Your proposal is the correct way to do things ^_^

GlassOfWhiskey:
Now let's proceed with the examination of the first observation.

**The copy fix is not so easy**. Let's start from the beginning. The goal of the task is to perform a simple copy from host fs to container fs. Docker has an appropriate command to do stuff like that. It is called `docker cp`. The problem is that, on Windows systems, `docker cp` **always create files with root user and group** inside the container. This means that, if `liferay` user needs to manipulate files (and it actually needs to) you have to perform a double copy step. In this case, you don't even need to tar and untar files. The solution adopted in my original npm plugin was the following:
`docker exec containerName mkdir -p temporaryPath bundleDirPath`
`docker cp hostPath containerName:temporaryPath`
`docker exec containerName /bin/cp -rf temporaryPath bundleDirPath`
`docker exec --user="root" containerName rm -rf temporaryPath`
This implies that you have to execute 4 commands for every deploy operation, which comes with poor time performances.

The `tar` pipe strategy, suggested in the [original documentation](https://docs.docker.com/engine/reference/commandline/cp/) of `docker cp` command, comes with a great performance improvement, because it can overcome the permission issues with a single command. Moreover, it can take advantage of NodeJS streams, which are very performant. Conversely, your solution has two drawbacks:

- it creates three child processes, which are quite heavy. Moreover, if you use `docker cp` explicitly, the `tar` step could be completely wasteful and maybe should be better to directly copy files inside the container

- it doesn't take into account the permission issue on Windows systems. I have to test it, but I suspect that the last `docker exec` fails on Windows systems if you don't add the `--user="root"` flag.

For these reasons, I would like to investigate a bit more if the `tar` pipe solution can be adapted for Liferay official containers.

I used Windows for my tests, but I suspect that the problem is not the host OS. Indeed, my Liferay containers are based on `centos:7`, while Liferay official containers are based on `alpine`. Surely this could imply that different `tar` versions are installed in the two environments. I will check a couple of things related to Docker images hierarchy and installed packages this evening, trying to find if there is a simple way to solve the issue.

If I won't be able to find a solution, or if a solution exists but is infeasible or excessively time consuming, I will reintroduce the double copy step method.

izaera:
Let me state that the problem cannot be with `tar` itself (in Alpine) because if it was that it wouldn't work by copying the file over. Unless, of course, `tar` behaves differently when reading the tar from command line (as opposed to fs). I didn't test that possibility, but it could be possible.

So, if it is not `tar`, what's the problem? I have no idea... It could be something in the middle: node.js, MacOS, docker, ...

Try to test it in Windows with the official images as you proposed and then I can retest your solution in my Mac and see if we can make it work in both.

Thanks :-)

izaera:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

izaera:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

izaera:
@GlassOfWhiskey I'm resending this PR with just your code and a commit of mine to show more detailed errors 👉 https://github.com/liferay/liferay-themes-sdk/pull/116

PR #114, opened 2018-11-29
Fixes #102 - Gulp watch fails on 7.1.1-GA2

### LINKED ISSUE #102: gulp watch doesn't work
I've just downloaded DXP 7.1.10-ga1 and I'm running it on linux
linux
with these properties on portal-ext:
portal-ext

These are the binaries/packages I have installed:
node-version
npm-version
gulp-version
generator-version

These are the steps I did:
- created a theme with the generator
yo
- executed `gulp watch`
- checked that the theme was deployed and active
gogo
- created a page with my custom theme
- changed `src/css/_custom.scss` and checked that `gulp watch` has detected the changes and tried to deploy
watch-file-changed

I don't see any errors, but the changes I make are never reflected on the portal.
(browser cache is disabled)

Additionally:
- If I use `gulp deploy` the theme is deployed.
- If I use `gulp deploy:gogo` the theme is NOT deployed, and I have this output:
deploy-gogo

izaera:


izaera:
Apparently the fix is not working. Will resend.

PR #153, opened 2019-02-26
Use Prettier to format all eligible files

wincent:
Closes: https://github.com/liferay/liferay-themes-sdk/issues/151


wincent:
I think I'm going to add an exclusion for the changelog because `github_changelog_generator` is going to keep producing a different result each time we run it.

julien:
Just started reviewing :)

PR #155, opened 2019-02-26
Remove rubySass

wincent:
We previously had [an issue](https://github.com/liferay/liferay-themes-sdk/issues/137) for removing the ruby-sass option from the theme generator because it is no longer required on Windows (which was the reason it was initially added).
    
In preparation for the next major release we remove the remaining traces of it. There is still one indirect reference to it in the form of a transitory dependency, as reported by `yarn why gulp-ruby-sass`:
    
```
=> Found "gulp-ruby-sass@2.1.1"
info Reasons this module exists
   - "_project_#liferay-theme-tasks#liferay-theme-deps-6.2" depends on it
   - Hoisted from "_project_#liferay-theme-tasks#liferay-theme-deps-6.2#gulp-ruby-sass"
```
    
This was added in 7b833e937500bc3e7339ec1b3eb24 ("Adds liferay-theme-deps-6.2 dependency to support upgrading from 6.2 with the 8.x version of the theme-tasks directly", August 2018). I am not totally sure, but I think we might be able to remove it from there too.
    
Test plan: `yarn test`
    
Closes: https://github.com/liferay/liferay-themes-sdk/issues/152

### DIFF
--- a/packages/generator-liferay-theme/generators/app/__tests__/index.test.js
+++ b/packages/generator-liferay-theme/generators/app/__tests__/index.test.js
@@ -315,7 +315,6 @@ describe('liferay-theme:app unit tests', function() {
 // 			function() {
 // 				var pkg = getPackage(defaults.themeId);
 //
-// 				chaiAssert.equal(pkg.liferayTheme.rubySass, false);
 // 				chaiAssert.equal(pkg.liferayTheme.templateLanguage, 'ftl');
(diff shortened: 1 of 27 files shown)


wincent:
@julien: this one builds on https://github.com/liferay/liferay-themes-sdk/pull/150 so I think I'll need to rebase it after that one lands.

julien:
Just started reviewing :)

PR #156, opened 2019-02-26
Remove update-notifier functionality

wincent:
We concluded that rather than fixing the reported bug where the update notifier was spuriously advising people to update even though they already had, that it was better just to remove the functionality. It is generally not urgent that people migrate to new versions, and posting a loud update warning may just end up desensitizing people to messages, or worse, making them anxious.

Closes: https://github.com/liferay/liferay-themes-sdk/issues/92


julien:
Just started reviewing :)

Issue #147, opened 2019-02-26
Add support for 7.2 Themes


jbalsas:
For the incoming 7.2 release of Liferay DXP, we should add support for 7.2 themes in the `generator-liferay-theme` project.

Subtasks:

- [ ] Create liferay-theme-deps-7.2 (or figure out how to remove that pattern)
- [ ] Offer 7.2 as a target choice

wincent:
Note: there was some discussion about whether these "liferay-theme-deps-7.x" packages are the right idea. On the one hand, they hide the info about the actual dependencies that are being used; on the other, they hide info about the actual dependencies that are being used. So it is convenient, but not 100% obvious that it's the best thing to do.

wincent:
Moved this in to the 9.0.0 milestone instead of 8.0.0-rc.4.

The implementation on the "master" branch was incomplete (see additional changes required in PR #187), testing was complicated due to the addition of the new-but-not-yet-published "liferay-theme-deps-7.2" package, and it is likely to cause confusion about which version to use in order to work with DXP 7.2 (it will be easier to just say "use v9 for DXP 7.2").

wincent:
Ok, this should be pretty much done at this point (via #187) so going to mark it as closed.

Ghadeerraa:
i need theme for liferay 7.2 ce ga2 
how i can get it ?

wincent:
@Ghadeerraa: I don't really understand the question. Are you talking about creating a theme from scratch? Installing a theme? Upgrading an a 7.1 theme?

The answer to all of these is likely to be found somewhere [on the help site](https://help.liferay.com/hc/en-us). For example, lots of info about creating 7.2 themes [here](https://help.liferay.com/hc/en-us/categories/360001751672):

![Screenshot 2020-05-12 -124937-KXX5ZEIZ@2x](https://user-images.githubusercontent.com/7074/81675389-0d054d80-944f-11ea-9f31-c66d955099f4.png)

Ghadeerraa:
i need liferay theme for liferay 7.2 CE GA2
can i have it ?

wincent:
Yeah, still not sure I understand the question. Liferay ships with a number of built-in themes (styled, unstyled, classic, admin); you don't have to do anything to "get" them as such, as they are already there. In 7.2 there are a number of sample themes as well (Fjord, Porygon, Westeros Bank) and which are mentioned in a number of tutorials I believe.

[...]

PR #158, opened 2019-02-26
Fixes #147 | Add support for 7.2 themes

julien:


wincent:
Looks ok to me. Going to merge and if I've missed anything we can address it in a follow-up.

PR #159, opened 2019-02-26
Remove lodash from liferay-theme-finder

wincent:
As we're not using Babel or polyfills in here at the moment, I tried to keep this pretty squarely within ES5-land; please let me know if I missed anything. There were some arrow functions in the code already on the test side, but I don't think I added any outside the tests. I am using a `Set` in one place, which is officially ES6; we could easily get away with just using an array there (we're doing intersection of tiny arrays).

Related: https://github.com/liferay/liferay-themes-sdk/issues/141

### DIFF
--- a/packages/liferay-theme-finder/index.js
+++ b/packages/liferay-theme-finder/index.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var _ = require('lodash');
 var async = require('async');
 var fs = require('fs-extra');
 var globby = require('globby');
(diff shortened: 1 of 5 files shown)


packages/liferay-theme-finder/index.js
```diff
@@ -9,17 +8,34 @@ var packageJson = require('package-json');
...
 
+function toArray(value) {
+	return Array.isArray(value) ? value : [value];
+}
+
+function intersection(a, b) {
+	a = toArray(a);
+	b = toArray(b);
+
+	const set = new Set(b);
```
wincent: This is the `Set` I mentioned in the PR summary.

julien:
Just started reviewing :)

packages/liferay-theme-finder/index.js
julien: Yes, I saw it.
I think that since this is "standard" ES6 it shouldn't be a problem

PR #163, opened 2019-02-27
Remove _.isPlainObject calls in divert.js (x2)

wincent:
Also worth noting: these two modules are almost exact duplicates. As a follow-up I am going to look at de-duping them.

Related: https://github.com/liferay/liferay-themes-sdk/issues/141


wincent:
https://github.com/liferay/liferay-themes-sdk/pull/165 is the follow-up to this, building on top of it (which, due to the way GitHub works, means that the commits are duplicated in that one).

julien:
Just started reviewing :)

PR #165, opened 2019-02-27
Fold "divert" into liferay-theme-tasks

wincent:
As noted in:

https://github.com/liferay/liferay-themes-sdk/pull/163

This function was duplicated almost verbatim in two packages. Centralize it in "liferay-theme-tasks" as the first step towards unmonorepoization discussed in:

https://github.com/liferay/liferay-themes-sdk/issues/164

Eventually, we might only have two packages:

- generator-liferay-theme (needs to be independent, so yeoman can find it).
- liferay-theme-tasks (effectively the public "API" of the SDK).

Test plan: `yarn test` and `yo ./packages/generator-liferay-theme` (works, apart from expected failure due to us not having published liferay-theme-deps-7.2 to NPM yet).


wincent:
Note: this one is built on top of https://github.com/liferay/liferay-themes-sdk/pull/163 so it starts with the same two commits. If we merge that one first, then I will rebase this one.

julien:
Just started reviewing :)

PR #167, opened 2019-02-27
Remove deprecated liferay-theme-finder package

wincent:
We're wanting to reduce complexity here, for the reasons stated in:

https://github.com/liferay/liferay-themes-sdk/issues/164

This package has 0 weekly downloads, and its functionality has already been folded in to "liferay-theme-tasks" in [liferay-theme-tasks/lib/theme_finder.js](https://github.com/liferay/liferay-themes-sdk/blob/84db0037fa76d8d6d405316a5790ac2e9f9111d9/packages/liferay-theme-tasks/lib/theme_finder.js).

Closes: https://github.com/liferay/liferay-themes-sdk/issues/166

PR #170, opened 2019-02-27
[RFC][WIP] Eliminate liferay-theme-deps-* packages

wincent:
This is just a WIP; there a bunch of failing tests because I haven't tackled the places where we have a dynamic resolution in place (where we look for "liferay-theme-deps-*" packages at runtime and use that to switch the "require" behavior).

Just pushing this as an early preview anyway. With this PR, the "packages/" dir contains:

- generator-liferay-theme
- liferay-theme-es2015-hook
- liferay-theme-mixins
- liferay-theme-tasks

Related: https://github.com/liferay/liferay-themes-sdk/issues/168

Disclaimer: built on top of #167 so it redundantly includes that commit as the first one in the chain.


packages/liferay-theme-tasks/README.md
```diff
@@ -258,12 +256,8 @@ Version of Liferay Portal this theme is intended for.
 
 ## Additional Theme Dependencies
 
-In order for themes to successfully build, they must declare additional dependencies in their package.json file.
+In order for themes to successfully build, they must declare additional dependencies such as [liferay-frontend-theme-unstyled](https://www.npmjs.com/package/liferay-frontend-theme-unstyled) and [liferay-frontend-theme-styled](https://www.npmjs.com/package/liferay-frontend-theme-styled) in their package.json file.
```
jbalsas: Will this always be the case? Shouldn't it suffice to add your base-theme as a dependency?

packages/liferay-theme-tasks/README.md
```diff
@@ -258,12 +256,8 @@ Version of Liferay Portal this theme is intended for.
...
 
-These additional dependencies are normally included via the [liferay-theme-deps-7.0](https://github.com/Robert-Frampton/liferay-theme-deps-7.0) and [liferay-theme-deps-6.2](https://github.com/Robert-Frampton/liferay-theme-deps-7.0) preset packages which are automatically added as dependencies to themes created with the [Theme Generator](https://github.com/liferay/generator-liferay-theme).
-
-These preset packages will include the necessary base themes/tools for compiling the theme, such as [liferay-frontend-theme-unstyled](https://www.npmjs.com/package/liferay-frontend-theme-unstyled) and [liferay-frontend-theme-styled](https://www.npmjs.com/package/liferay-frontend-theme-styled).
-
-If you would like to implement specific versions of these dependencies, you can specify them as dependencies directly in your theme's [package.json file](https://gist.github.com/Robert-Frampton/86a084704afe9bca3e3ccc05d2709a5b).
+These additional dependencies are automatically to themes created with the [Theme Generator](https://github.com/liferay/generator-liferay-theme).
```
jbalsas: nit: Possible typo, unnecessary `to themes`?

packages/liferay-theme-tasks/README.md
wincent: At the moment I think they're still required (because even though you might extend one of those themes, it won't get installed for you unless its in the package.json). You don't have to add them yourself though; the generator does it.

packages/liferay-theme-tasks/lib/dependencies.js
```diff
@@ -0,0 +1,43 @@
...
+	'liferay-frontend-theme-unstyled': '3.0.13',
+});
+
+/**
+ * Dependencies for each version of Liferay. Currently, uniform across versions
+ * 7.0 through 7.2.
+ */
+const DEPENDENCIES = {
+	'BASE': BASE_DEPENDENCIES,
+	'7.0': BASE_DEPENDENCIES,
```
jbalsas: that a `7.0` theme would have the same `styled` and `unstyled` deps sounds sketchy... maybe we had an issue already, but usually, those themes will be released from the appropriate maintenance branch, so I'd expect a family of `2.x` themes for `7.0`, `3.x` for `7.1` and `4.x` once the branch for `7.2` is cut

jbalsas: Yeah, it seems it should've been sth like that from https://github.com/liferay/liferay-themes-sdk/pull/170/files#diff-c6b182268a8860d12a1abb1486a8149bL17?

packages/liferay-theme-tasks/README.md
wincent: Good catch.

packages/liferay-theme-tasks/lib/dependencies.js
wincent: Ah yes, you're right. I noticed this but then forgot it along the way. This is what we used to have prior to this PR:

7.0:

```
               "compass-mixins": "0.12.10",
               "gulp-sass": "3.2.0",
               "liferay-frontend-common-css": "~1.0.4",
               "liferay-frontend-theme-classic-web": "~2.0.0",
               "liferay-frontend-theme-styled": "~2.0.0",
               "liferay-frontend-theme-unstyled": "~2.1.0",
               "liferay-theme-deps-normalize": "^8.0.0-rc.3"
```

7.1:

```
               "compass-mixins": "0.12.10",
               "gulp-sass": "3.2.0",
               "liferay-frontend-common-css": "1.0.4",
               "liferay-frontend-theme-classic-web": "2.0.2",
               "liferay-frontend-theme-styled": "3.0.13",
               "liferay-frontend-theme-unstyled": "3.0.13",
               "liferay-theme-deps-normalize": "^8.0.0-rc.3"
```

7.2:

```
               "compass-mixins": "0.12.10",
               "gulp-sass": "3.2.0",
               "liferay-frontend-common-css": "1.0.4",
               "liferay-frontend-theme-classic-web": "2.0.2",
               "liferay-frontend-theme-styled": "3.0.13",
               "liferay-frontend-theme-unstyled": "3.0.13",
               "liferay-theme-deps-normalize": "^8.0.0-rc.3"
```

jbalsas: So, the tricky thing here is... should we choose fixed or moving versions to define the dependencies?

packages/liferay-theme-tasks/README.md
jbalsas: Yeah, I meant that if you just added `styled`... and `styled` properly defined its dependency on `unstyled`, then you wouldn't need to specify both, right?

Same way, if you eventually extended from another theme... and it depended on `styled` and so on... you'd just need to list the `base`.

My gut tells me that `styled` and `unstyled` aren't conventional themes and so they need to be treated differently, so I'm just putting this out here so we can follow up, nothing more :)

packages/liferay-theme-tasks/lib/dependencies.js
wincent: I created https://github.com/liferay/liferay-themes-sdk/issues/171 to make sure we don't forget to bump the versions once the v4.0.x releases of the themes have been cut.

packages/liferay-theme-tasks/README.md
wincent: Yeah, just to document this: the styled theme definitely doesn't currently declare an (NPM-level) dependency on unstyled. And looking in the two relevant places (which, as far as I can tell, are [liferay/com-liferay-frontend-theme](https://github.com/liferay/com-liferay-frontend-theme) and [the module/apps/frontend-theme-* directories](https://github.com/liferay/liferay-portal/tree/6eeea9163a7864817ecb91f49162f95000ead88f/modules/apps) in the liferay-portal repo, the only extension relationships we have defined are:

- frontend-theme-classic and frontend-theme-admin have "styled" as a base theme.
- frontend-theme-fjord, frontend-theme-westeros-bank, frontend-theme-porygon also extend  "styled".

I haven't yet looked in detail at the gulp tasks to see how build and deploying works; given that styled is bundled inside the portal, I don't know under what circumstances, if any, the build/deploy will use that instead of relying on a local copy specified in the package.json dependencies.

wincent: For anybody stumbling across this in the future: the [liferay/com-liferay-frontend-theme repo](https://github.com/liferay/com-liferay-frontend-theme) is effectively legacy and can be ignored.

wincent:
I'm going to re-push this as a draft.

Issue #172, opened 2019-02-28
Remove generator dependence on globally installed gulp instance


wincent:
If you try to generate a theme without a globally installed copy of `gulp` you'll get `Error: spawn gulp ENOENT` at then end. We do tell people (in the README) to install `gulp` globally, but that shouldn't be necessary: it is already a dependency of the generator and we can just use the generator's copy.

wincent:
Fix is up in #173. Note that I labeled this as "bug" even though it was technically working as intended, because I think users see an error message like `Error: spawn gulp ENOENT` and perceive it as a bug.

PR #173, opened 2019-02-28
Remove generator dependence on globally installed gulp instance (#172)

wincent:
If you try to generate a theme without a globally installed copy of gulp you'll get `Error: spawn gulp ENOENT` at the end. We do tell people (in the README) to install gulp globally, but that shouldn't be necessary: it is already a dependency of the generator and we can just use the generator's copy.

Test plan: `yo ./packages/generator-liferay-theme` without a globally installed copy of `gulp` and see it work without blowing up.

Closes: https://github.com/liferay/liferay-themes-sdk/issues/172

### DIFF
--- a/packages/generator-liferay-theme/README.md
+++ b/packages/generator-liferay-theme/README.md
@@ -6,8 +6,8 @@
 
 ## Dependencies
 
-1. Install [Node.JS](http://nodejs.org/), if you don't have it yet.<br>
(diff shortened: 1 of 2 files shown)


julien:
Just started reviewing :)

wincent:
Checked with @julien about this one over Slack and got the green light to merge it.

PR #174, opened 2019-02-28
[RFC] Eliminate liferay-theme-deps-* packages

wincent:
This is just a WIP draft; there a bunch of failing tests because I haven't tackled the places where we have a dynamic resolution in place (where we look for "liferay-theme-deps-*" packages at runtime and use that to switch the "require" behavior).

Just pushing this as an early preview anyway. With this PR, the "packages/" dir contains:

- generator-liferay-theme
- liferay-theme-es2015-hook
- liferay-theme-mixins
- liferay-theme-tasks

Related: https://github.com/liferay/liferay-themes-sdk/issues/168

I pushed an earlier version of this as #170, but didn't mark it as a draft, which might have caused some confusion, so re-pushing now as a draft. Will be adding more commits on top to fix the remaining issues.


wincent:
Dropping this one. It is superseded by: https://github.com/liferay/liferay-js-themes-toolkit/pull/187

PR #176, opened 2019-02-28
Remove unnecessary dynamism of getDependencyName

wincent:
We go a combination of `divert` (which dynamically selects a version-specific module) and the exported `getDependencyName` functions to generated what is essentially a fixed set of names and which is unlikely to ever change. As such it is unnecessary complexity that we can always add back in if we ever need it, but I think it's unlikely we ever will.

Extracted this out as a relatively uncontroversial bit from:

https://github.com/liferay/liferay-themes-sdk/pull/174

Test plan: `yarn test`


wincent:
This is another one I need to build on top of so going to merge it and will follow-up to address any feedback.

PR #177, opened 2019-03-01
Freshen lockfile

wincent:


wincent:
Don't know why it's not up to date. #167, which went in recently, did update the lockfile. 🤷‍♂️

PR #178, opened 2019-03-01
Overhaul normalize processing

wincent:
In this PR you can see a couple of conservative improvements I made to the "normalization":

- Removing unnecessary parametrization.
- Make tag injection idempotent (arguably fixing a latent bug).

followed by a more aggressive change, which is completely deleting the "liferay-theme-deps-normalize" plug-in and making it so that we don't mutate installed node_modules any more. From the commit message:

This commit deletes the liferay-theme-deps-normalize package from the repo and replaces the hack that it used to use (mutating the installed liferay-frontend-theme-unstyled packaged in node_modules) with a build transform that injects the tags as the files are written into the build directory.

Test plan:

Unfortunately, it's complicated:

- `yarn test`
- In packages/liferay-theme-tasks/package.json, remove dependency on liferay-theme-deps-7.2 because it is not yet published and will otherwise cause things to blow up.
- Use the local copy of the generator to create a new theme called "test"; ie. `yo ./packages/generator-liferay-theme`
- `cd test-theme` and change this line in gulpfile.js to make it use the local tasks package:

```diff
- var liferayThemeTasks = require('liferay-theme-tasks');
+ var liferayThemeTasks = require('../packages/liferay-theme-tasks');
```

- Run the "build" task; ie. `node_modules/.bin/gulp build`
- Inspect build/templates/portal_normal.ftl and see the tags have been    injected.

One disclaimer about the `injectJS` function: I wrote code to handle both the `isStream()` and `isBuffer()` case, but `console.log` shows that we only get buffers so one of the branches is effectively untested. If there is a bug in there and it blows up, we'll have to deal with that, but it's possible we'll never get a stream here. FWIW, I based what I wrote in that branch on these docs:

- https://gulpjs.org/writing-a-plugin/
- https://gulpjs.org/writing-a-plugin/dealing-with-streams.html
- https://www.npmjs.com/package/through2
- https://nodejs.org/api/stream.html#stream_transform_transform_chunk_encoding_callback
- https://nodejs.org/api/stream.html#stream_transform_flush_callback


wincent:
Was originally trying to extract this from #174 but ending up going a different route.

wincent:
Going to merge this one in order to make #179 easier to review. I can follow-up on any feedback related to this one.

PR #179, opened 2019-03-01
Replace the "divert()" abstraction with something simpler

wincent:
Simplicity is in the eye of the beholder, of course, but the main value prop here is that you can see what the differences are between each version by looking in a single file instead of having to search across the file-system.

Details are in the individual commit messages. Seems to work pretty well in my testing. Some of the names could perhaps use a tweak.

Originally was going to extract some stuff from https://github.com/liferay/liferay-themes-sdk/pull/174 but ended up going down this different path.


wincent:
Ah one other thing; builds on https://github.com/liferay/liferay-themes-sdk/pull/178 — which I might just merge to make this all simpler to review.

packages/liferay-theme-tasks/lib/lookup.js
```diff
@@ -0,0 +1,31 @@
...
+	return {
+		'baseThemeDependencies': require('./lookup/base').getBaseDependencies,
+		'devDependencies': require('./lookup/dependencies').devDependencies,
+		'kickstart:choices': require('./lookup/kickstart').choices,
+		'kickstart:afterPromptThemeSource': require('./lookup/kickstart')
+			.afterPromptThemeSource,
+		'template:choices': require('./lookup/template').choices,
+		'template:isLanguage': require('./lookup/template').isLanguage,
+		'template:printWarnings': require('./lookup/template').printWarnings,
+	}[key](version);
```
jbalsas: If I understand this... now every call to `lookup` will actually involve resolving and running the code for every version-aware code regardless of wheter it was asked or not.

Not saying we should change this, but maybe we could do something like:

```
const [lookupKey, lookupProp] = key.split(':');

return require(`./lookup/${lookupKey}`)[lookupProp];
```

We'd need to find something for `baseThemeDependencies` and `devDependencies` to match the pattern... 🤷‍♂️

wincent: Correct. Every time you call `lookup` we invoke `require`; the module body is only evaluated once, the functions *are* called with the `version` parameter every time. I didn't bother with trying to cache the result because it would likely add complexity in an area where performance is not a concern and as you know, cache invalidation is hard.

I think we should avoid dynamically constructing `require` statements because they will defeat static analysis by TypeScript (or any other similar tool we consider switching to).

jbalsas: 👍

julien:
Just started reviewing :)

packages/liferay-theme-tasks/lib/lookup/dependencies.js
```diff
@@ -0,0 +1,20 @@
...
+	const dependencies = [
+		`\t"gulp": "3.9.1",`,
+		`\t\t"liferay-theme-tasks": "8.0.0-rc.3",`,
+	];
+
+	if (version === '7.0') {
+		dependencies.push(`\t\t"liferay-theme-deps-7.0": "8.0.0-rc.3"`);
+	} else if (version === '7.1') {
+		dependencies.push(`\t\t"liferay-theme-deps-7.1": "8.0.0-rc.3"`);
+	} else if (version === '7.0') {
```
julien: I know we'll probaly want to change this at some point, but I think we meant 

``` } else if (version === '7.2') { ```

Here, otherwise npm will fail parsing the generated `package.json`

julien: nevermind this is an outdated comment

PR #180, opened 2019-03-04
Don't use string-concatenation to build up devDependencies

wincent:
When I ported us from `divert()` to `lookup()` in #179 I didn't change the string concatenation because I didn't want to change too many things at once.

Here as a follow-up I'm replacing it with `JSON.stringify()`. I'm not totally convinced by this change though. While we got rid of the concatenation (yay!), we now have this `split()`/`join()`/`replace()` song and dance to get the outputted JSON looking exactly how we want it. If we didn't care about that we could just leave all that off.

Anyway, going to put this up as an RFC anyway and see.

Test plan: `yarn test`, use `yo ./packages/generator-liferay-theme` to create a new theme and then inspect that the package.json is correct, including the tabs.

sample


julien:
Just started reviewing :)

PR #181, opened 2019-03-05
Update eslint and eslint-config-liferay to latest

wincent:
See the individual commit messages for details, but in essence this is just updating the dependencies, adding a `.prettierrc.json` (because the new eslint wants to reformat everything), fixing lints that popped up, and turning off suppressions (and fixing the lints that popped up as a result).

Related: https://github.com/liferay/eslint-config-liferay/issues/21

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,4 @@
-{
+module.exports = {
     "env": {
         "node": true,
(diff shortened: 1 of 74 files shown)


julien:
Just started reviewing :)

PR #186, opened 2019-03-07
Rename SDK reference to "toolkit"

wincent:
If we end up actually relocating the repo itself, there are a bunch of URLs that we should probably update too, as per:

https://help.github.com/en/articles/renaming-a-repository

Related: https://github.com/liferay/liferay-themes-sdk/issues/185


julien:
Just started reviewing :)

Issue #184, opened 2019-03-06
Clarify "compatibility matrix" between versions of liferay-js-themes-toolkit and Liferay DXP


wincent:
At the moment the current release (v8.0.0-rc.3) provides varying degrees of support for different versions of Liferay DXP; for example:

- You can create new themes for Liferay DXP 7.0. or 7.1 (and on the unreleased "master" branch, you can also create 7.2 themes, which are basically the same as 7.1 themes but depending on "liferay-theme-deps-7.2" instead of "liferay-theme-deps-7.1").
- There are upgrade tasks for going from 6.2 to 7.0, and from 7.0 to 7.1.
- "liferay-theme-tasks" offers a huge number of tasks that can be used inside themes, but unfortunately it builds on Gulp v3 which currently has a lot of insecure dependencies (as identified by `npm audit`/`yarn audit`) and is API-incompatible with Gulp v4.

With the release of DXP 7.2 our "compatibility" matrix becomes more complicated because, if we decide to continue down the same path we've been on so far, all of those "liferay-theme-tasks" will now have to be aware of *four* versions of DXP (6.2, 7.0, 7.1, 7.2) and we'll additionally need a new 7.1-to-7.2 upgrade functionality.

What I'd like to propose here is the following:

- Draw a line in the sand and cut v8.0.0 final from the existing v8.0.0-rc3 base; this version will support the features listed at the top (ability to upgrade a theme from 7.0 to 7.1, ability to create themes for 7.0 and 7.2 etc). Call this the "8.x-stable" branch and only apply bugfixes to it.
- On "master", prepare a new v9.x series which will drop support for some of the older features which are causing maintenance overhead; for example:
  - v9 would be able to:
    - Create themes for DXP 7.2 (only!).
    - Upgrade a DXP 7.1 theme to DXP 7.2.
  - v9 would *not* be able to:
    - Create themes for 7.0 or 7.1.
    - Upgrade themes from 6.2 to 7.0, or from 7.0 to 7.1.
  - Additionally, v9 will work exclusively with Gulp v4 (we *may* be able to provide some kind of shim to expose the tasks to somebody running Gulp v3, but I am not sure if that is feasible).

Basically, the idea would be that if you want to create a new theme from scratch, you download v9 of "generator-liferay-theme" and use it. If you want to upgrade a 7.1 theme to 7.2, you update you "liferay-theme-tasks" dependency to v9 and use it.

[...]

PR #187, opened 2019-03-07
Prepare for v8/v9 split

wincent:
This is a draft/RFC at this point, based on what I suggested in: https://github.com/liferay/liferay-themes-sdk/issues/184

I'll add more commits to this as I go. Some of the commits in here could be split off and merged independently of whether we decide to what #184 proposes, but for now I am just going to accumulate changes here so as to keep the number of PRs manageable.

These are targeting the "master" branch. I created an "8.x-stable" branch that won't hurt to have around in any case.


packages/liferay-theme-tasks/lib/upgrade/7.0/upgrade.js
```diff
@@ -73,11 +73,11 @@ module.exports = function(options) {
 					patterns: [
 						{
 							match: /(DTD Look and Feel )\d(?:\.\d+)+(\/\/EN)/g,
-							replacement: '$17.0.0$2',
+							replacement: '$17.1.0$2',
 						},
 						{
 							match: /(liferay-look-and-feel_)\d(?:_\d+)+(\.dtd)/g,
-							replacement: '$17_0_0$2',
+							replacement: '$17_1_0$2',
```
wincent: This is the bit I was referring to in the commit message where I said we weren't actually updating the version numbers.

The 7.1 DTD actually exists, as you can see here: https://github.com/liferay/liferay-portal/blob/master/definitions/liferay-look-and-feel_7_1_0.dtd

wincent:
Still a work-in-progress, but adding you as a reviewer @julien in case you want to take a sneak peak at how this is evolving.

julien:
@wincent I definetly will have a look, thanks

wincent:
Still more that can be done here but removing "draft" status because I think it's pretty close now.

wincent:
Just adding a reminder here for myself that https://github.com/liferay/liferay-js-themes-toolkit/pull/206 removes 7.2 support from the master branch, which means that when I merge master into this PR I'll need to remember to add back the relevant bits (I'll get merge conflicts anyway, but just in case).

Issue #127, opened 2018-12-19
Make `watch` task more robust


jbalsas:
Our `watch` task currently depends on the `gogo` shell being available and enabled in a running instance of Liferay Portal. This has proven to be a bit unstable and unreliable over time. 

We should try to investigate new alternatives or ways to provide this that don't necessarily depend on deploying bundles to Liferay.

One option could be to have some type of dev-server (webpack or standalone) so we could simply proxy the updated css files on request

wincent:
Going to close this now that #190 and #201 are in, along with #209.

PR #188, opened 2019-03-07
Research: Make `watch` task more robust (#127)

julien:
Here are some changes I made for the watch task to use 
`webpack-dev-server` instead of `browser-sync`

I think it would be good to start a conversation on this topic

- The first thing I'd like to mention is that is pretty complicated to test the generator.
We should try to make this easier if it's possible, but generating a theme correctly with the local versions of these packages is extremly complicated and not documented.

- Using `weback-dev-server` seemed like a good idea at first because it can be used as a server and a proxy (`browser-sync` can only do one or the other but not both at the same time), but it feels like we are duplicating the existing work, because we already have gulp tasks to compile the sass files and we're generating a webback configuration file that does the same thing.

- What I haven't been able to do is having live reload without injecting a script tag in the templates.
This is probably due to something badly configured in the webpack configuration file or something 
missing but I've looked at all the options and haven't found the correct way to do it.
[EDIT]: One thing I have yet to try is setting the `publicPath` (under `devServer.proxy`) to the actual URL of the dev server (probably localhost:9080) 

- The other main problem is that there's no way to know when the deploy task was actually finished (on the server) and reload the browser at the correct moment.

So I'm leaving this as is for now, if someone wants to have a look at it or propose changes I'd be more than happy

### DIFF
--- a/packages/generator-liferay-theme/generators/app/index.js
+++ b/packages/generator-liferay-theme/generators/app/index.js
@@ -8,7 +8,7 @@ var path = require('path');
 var yeoman = require('yeoman-generator');
 var yosay = require('yosay');
 
-var lookup = require('liferay-theme-tasks/lib/lookup');
+var lookup = require('../../../liferay-theme-tasks/lib/lookup');
 
 module.exports = yeoman.generators.Base.extend({
 	initializing() {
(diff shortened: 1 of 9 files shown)


trytuna:
When we expect that this is going to be merged and releases? As far as I know there is no workaround for that issue?

wincent:
@trytuna: it doesn't exactly work in its current form, so this can't be merged as-is. We're still trying to figure out the best way to make this work, but we'll update here once we know more.

wincent:
See also https://github.com/liferay/liferay-js-themes-toolkit/pull/190 which is an alternative approach.

wincent:
Spring cleaning this one because #190 is going to go in.

PR #190, opened 2019-03-11
Make `watch` task more robust (2nd approach)

julien:
This is another test using `http-proxy` and `tinylr` for live reloading.

Please tell me what you think when you have a chance.

Thanks!

### DIFF
--- a/packages/liferay-theme-tasks/tasks/deploy.js
+++ b/packages/liferay-theme-tasks/tasks/deploy.js
@@ -188,6 +188,10 @@ function registerTasks(options) {
 	function reloadBrowser(fileGlobs) {
 		const browserSync = gulp.browserSync;
 
+		if (!browserSync) {
+			return;
+		}
+
 		if (fileGlobs) {
 			browserSync.reload(fileGlobs);
 		} else {
--- a/packages/liferay-theme-tasks/tasks/watch.js
+++ b/packages/liferay-theme-tasks/tasks/watch.js
@@ -63,8 +63,8 @@ module.exports = function(options) {
 				portfinder.getPortPromise({port: 9080}),
 				portfinder.getPortPromise({port: 35729}),
 			]).then(([httpPort, tinylrPort]) => {
-					storage.set('webBundleDir', 'watching');
-					startWatch(httpPort, tinylrPort, proxyUrl);
+				storage.set('webBundleDir', 'watching');
+				startWatch(httpPort, tinylrPort, proxyUrl);
 			});
 		});
 
@@ -85,7 +85,6 @@ module.exports = function(options) {
 	gulp.task('watch:osgi:clean', function(cb) {
 		cb();
 	});
-
 
 	/**
 	 * Clean the remote exploded build dir in docker
@@ -138,6 +137,8 @@ module.exports = function(options) {
 		runSequence.apply(this, taskArray);
 	});
 
+	let livereload;
+
 	gulp.task('watch:reload', function(cb) {
 		const changedFile = storage.get('changedFile');
 		const srcPath = path.relative(process.cwd(), changedFile.path);
@@ -152,24 +153,6 @@ module.exports = function(options) {
 		cb();
 	});
 
-	gulp.task('watch:reload:slow', function(cb) {
-		const changedFile = storage.get('changedFile');
-		const srcPath = path.relative(process.cwd(), changedFile.path);
-		const dstPath = srcPath.replace(/^src\//, '');
-		const urlPath = `${resourcePrefix}/${distName}/${dstPath}`;
-
-		setTimeout(() => {
-			livereload.changed({
-				body: {
-					files: [urlPath],
-				},
-			});
-			cb();
-		}, 10000);
-	});
-
-	let livereload;
-
 	/**
 	 * Start live reload server and watch for changes in project files.
 	 * @param {int} httpPort   The port for the http server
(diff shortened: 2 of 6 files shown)


wincent:
Awesome that you found another way!

packages/liferay-theme-tasks/tasks/deploy.js
```diff
@@ -188,6 +188,10 @@ function registerTasks(options) {
 	function reloadBrowser(fileGlobs) {
 		const browserSync = gulp.browserSync;
 
+		if (!browserSync) {
```
julien: Note: this is temporary until I remove `browserSync` from the other tasks, sorry for leaving that in

packages/liferay-theme-tasks/tasks/watch.js
```diff
@@ -1,38 +1,33 @@
...
 const portfinder = require('portfinder');
-
 portfinder.basePort = 9080;
 
-const CONNECT_PARAMS = {
-	port: 11311,
-};
+const DEPLOYMENT_STRATEGIES = themeUtil.DEPLOYMENT_STRATEGIES;
 const EXPLODED_BUILD_DIR_NAME = '.web_bundle_build';
+const IGNORED_EXTENSIONS = ['.ftl', '.vm'];
```
jbalsas: I think `.tpl` might still be around for layout templates

packages/liferay-theme-tasks/tasks/watch.js
```diff
@@ -170,36 +146,74 @@ module.exports = function(options) {
 		runSequence.apply(this, taskArray);
 	});
 
+	gulp.task('watch:reload', function(cb) {
+		const changedFile = storage.get('changedFile');
+		const srcPath = path.relative(process.cwd(), changedFile.path);
+		const dstPath = srcPath.replace(/^src\//, '');
+		const urlPath = `/o/${distName}/${dstPath}`;
```
jbalsas: The `/o` prefix might be problematic because it can be changed...  at the very least, this should be configurable, I think.

packages/liferay-theme-tasks/tasks/watch.js
```diff
@@ -170,36 +146,74 @@ module.exports = function(options) {
...
+			});
+
+			if (req.url === '/') {
+				// Inject the livereload script
+				res.write(`<script>document.write('<script src=\"http://localhost:${TINYLR_PORT}/livereload.js?snipver=1\"></' + 'script>')</script>`);
+			}
+
+			if (req.url.match(themePattern)) {
+				const parsedUrl = url.parse(req.url);
+				const pathname = parsedUrl.pathname.replace(`/o/${distName}/`, '');
```
jbalsas: This is a bit confusing... we're matching against a regex, and then here, rather than having maybe executed the regex and use a capturing group are removing an expected prefix... seems a bit of dual work, regex on one side, urls on the other... 🤔

packages/liferay-theme-tasks/tasks/watch.js
julien: Thanks, I remember you mentioned that.

packages/liferay-theme-tasks/tasks/watch.js
julien: Ok, I will think about a good way to make this configurable

packages/liferay-theme-tasks/tasks/watch.js
julien: Yes, I agree. I'll think of a smarter way to do this.

trytuna:
The redeployment is working fine but the reload is not happening. I dug a bit in the code and figured out that the reload event is sent to the browser but is not properly executed.

I fixed that by editing `./node_modules/liferay-theme-tasks/node_modules/livereload-js/dist/livereload.js` on line 409 with `window.location.reload();`.

It seems that this is a bug on livereload-js. I am not 100% sure if that's true so it would be great if the author of this PR could look more into that.

Thanks

julien:
@trytuna thanks for the comment.

I wanted to ask you more details about it: 

> the reload event is sent to the browser but is not properly executed.

Does this happen for `scss` and `js` files?

I also wanted to let you know that is still "work in progress"; so for example, right now if a template is changed it the browser will not be reloaded (yet).

trytuna:
> Does this happen for `scss` and `js` files?

I've tested it with scss files

julien:
@trytuna ok thanks, I'll have a look.
Can I also ask you which OS and browser you're using?

trytuna:
> @trytuna ok thanks, I'll have a look.
> Can I also ask you which OS and browser you're using?

Sure

macOS 10.14.3 (18D109)
Google Chrome 72.0.3626.121

julien:
@trytuna thanks for the info, I'm currently using the same OS (a "bit" older - macOS 10.11) and browser (same version) and can't seem to reproduce the issue here, do you have any extensions enabled that could be  (ad-blocker?)

trytuna:
> @trytuna thanks for the info, I'm currently using the same OS (a "bit" older - macOS 10.11) and browser (same version) and can't seem to reproduce the issue here, do you have any extensions enabled that could be interfering (ad-blocker?)

I tried that again with Chrome in incognito mode and all plugins disabled. I've also tried Safari and Firefox. Same Issue :(.

A bit more details on my setup:

    gulp -v
    [10:14:57] CLI version 2.0.1
    [10:14:57] Local version 3.9.1

    I am using Liferay 7.1.0-ga1 and generator-liferay-theme@8.0.0-rc.3 for generating my theme.

I try to dig a bit deeper to get a better glimpse of the problem.

julien:
@trytuna thanks for the update, I will have a look at that shortly

julien:
@wincent @jbalsas a quick update: reloading the assets (css, images, javascript) works better right now, I still have to check what @trytuna mentioned.

Concerning the templates I'm really not sure how to approach the "reload", but if you have suggestions I'm all ears.

Thanks

trytuna:
Following up on my last comment.

The issue I think is that the reload in the browser happens before Liferay is done deploying. An approach would be to trigger an event from Liferay and send it to the gulp watch task.

Issue #97, opened 2018-09-28
Upgrade default dependencies


chrberndt:
Currently (liferay-theme-tasks": "8.0.0-beta.5) a lot of dependencies of a newly created theme are outdated: 

```
...
added 972 packages from 540 contributors and audited 10389 packages in 17.807s
found 43 vulnerabilities (21 low, 22 high)
  run `npm audit fix` to fix them, or `npm audit` for details
[12:12:13] Using gulpfile ~/liferay/my-angular-app-theme/gulpfile.js
[12:12:13] Starting 'plugin:init'...
? Select your deployment strategy Local App Server
...

```
and most of them can't be updated automatically: 

```
...
berndt@x270:~/liferay/yet-another-liferay-theme$ npm audit fix
npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN yet-another-liferay-theme@1.0.0 No description
npm WARN yet-another-liferay-theme@1.0.0 No repository field.
npm WARN yet-another-liferay-theme@1.0.0 No license field.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.4: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

added 2 packages from 1 contributor and updated 2 packages in 3.907s
fixed 3 of 43 vulnerabilities in 10389 scanned packages
  35 vulnerabilities required manual review and could not be updated
  1 package update for 5 vulns involved breaking changes
  (use `npm audit fix --force` to install breaking changes; or refer to `npm audit` for steps to fix these manually)
...
```

shimy1984:
I see this vulnerability warning as well. Looks like it has to do with some gulp plugins that are in use.

wincent:
I created https://github.com/liferay/liferay-themes-sdk/issues/148 specifically for the Gulp v4 upgrade because that is going to be a rather large change.

wincent:
See also https://github.com/liferay/liferay-themes-sdk/issues/141 which is about the audit warnings related to Lodash. Keeping that one as a separate ticket because it might be simpler (or rather, better) to just remove Lodash (not sure though as I haven't yet surveyed to see how extensively we use it).

wincent:
[...]

PR #192, opened 2019-03-12
Update yeoman-generator v0.18.0 to 3.2.0 (for v9 of toolkit)

wincent:
Major changes between versions:

- You `extend` the base `Generator` class now instead of using objects.
- This in turn means that the `_.cloneDeep` and `_.merge` stuff that we were doing now to base the layout and themelet generators off of the main "app" generator no longer works. As an alternative, I extracted out the sub-methods that were in the "priority" objects and turned them into methods that can be called explicitly.
- Furthermore, Yeoman won't fall through to using a superclass version of a "priority" method like `configuring()` or `prompting()` so you have to call `super...()` instead.
- `this.template()` was removed, requiring us to move to `this.fs.copyTpl()` instead, which requires us to explicitly use `this.templatePath()` and `this.destinationPath()`.
- `this.installDependencies()` no longer takes a callback argument, requiring me to listen for an `npmInstall:end` event explicitly.
- I got rid of the `_yosay` property because the logical analog, ES6 class fields, are still only a proposal -- detailed here: https://github.com/tc39/proposal-class-fields -- and instead just used `this.options.namespace` which produces ever-so-slightly different output (ie. "liferay-theme" instead of "Liferay Theme", "liferay-theme:layout" instead of "Liferay Layout Template" etc).
- Need to change our interpolation sigil in a `_package.json` template otherwise the quotes end up getting HTML-escaped, producing invalid JSON.

Basically all seems to work with the only truly ghastly hack being the song-and-dance required to support an "Any" version specifier in themelets. We can't let the superclass see "`Any`" or "`*`" as a version because it will explode when it calls `lookup('devDependencies')`, so we temporarily set 7.2 as the version and then later replace it with "*". As a follow-up I can probably look for a nicer way.

After this update, our `yarn audit` report drops from:

    106 vulnerabilities found - Packages audited: 145078
    Severity: 36 Low | 29 Moderate | 41 High

to:

    77 vulnerabilities found - Packages audited: 141898
    Severity: 26 Low | 19 Moderate | 32 High

Test plan:

```
yo ./packages/generator-liferay-theme
yo ./packages/generator-liferay-theme/generators/layout
yo ./packages/generator-liferay-theme/generators/themelet
```

Also, `yarn test`; not green yet but I'll investigate that presently and squash something into this to fix it.

Related: https://github.com/liferay/liferay-js-themes-toolkit/issues/97

Currently basing this one off of https://github.com/liferay/liferay-js-themes-toolkit/pull/187 — depending on the fate of that one may need to backport this on top of "master" instead, but now that I've done this once that wouldn't be *too* bad.


packages/generator-liferay-theme/package.json
```diff
@@ -41,7 +41,7 @@
 		"lodash": "^4.17.10",
 		"lodash-bindright": "^1.0.1",
 		"minimist": "^1.2.0",
-		"yeoman-generator": "^0.18.0",
+		"yeoman-generator": "^3.2.0",
```
jbalsas: 0.18.0 -> 3.2.0... 😱

wincent: Just a tiny update.

wincent:
This was supposed to be a draft PR, BTW, but I refreshed the page before submitting and apparently the setting is not "sticky".

wincent:
Going to make a re-roll of this for the (current) "master" (soon, "8.x") branch. This one will still apply for v9 though so keeping it open.

wincent:
Ok, I am going to merge this into the "wincent/compatibility" branch (the one corresponding to #187) so that we just have one branch/PR to juggle for this major change.

Issue #193, opened 2019-03-12
Fold liferay-plugin-node-tasks into liferay-theme-tasks


wincent:
This package:

https://www.npmjs.com/package/liferay-plugin-node-tasks

is only used by "liferay-theme-tasks" and "liferay-theme-tasks-themelets" according to the "Dependents" tab on that NPM page.

We should fold it into this repo and concretely, into the "liferay-theme-tasks" package, to clear up the mysterious and fragile dependency relationship between the two. Additionally, this will make upgrading to Gulp v4 (#148) more straightforward, because we can do it all in one go instead of trying to coordinate an upgrade across two packages.

As best as I can tell, the way things currently are is as follows:

- "liferay-theme-tasks" depends on "liferay-plugin-node-tasks".
- Both register a "build" task, so one clobbers the other, I think (concretely, I think "liferay-theme-tasks" registers the tasks from "liferay-plugin-node-tasks" and then overwrites the latter's "build" with its own).
- "liferay-theme-tasks" ends up calling "liferay-plugin-node-tasks" in a couple of places; for example, "deploy:gogo" will invoke "plugin:deploy-gogo", "deploy:war" will invoke "plugin:deploy" etc.
- Those plugin tasks depend on "build", but due to the overwriting mentioned above, I *think* they actually end up calling the theme tasks version of "build", which has already run at that point, and at least under Gulp v3 (I *think*) won't run again.
- https://www.npmjs.com/package/liferay-plugin-node-tasks claims that "liferay-plugin-node-tasks" lives at: https://github.com/Robert-Frampton/liferay-plugin-node-tasks but it actually lives at: https://github.com/jbalsas/liferay-plugin-node-tasks (we should correct that link in the next release and ideally move this into the "liferay" GitHub org).

Related internal ticket: https://issues.liferay.com/browse/IFI-553

PR #194, opened 2019-03-13
Fix post-install for layout generator

wincent:
In testing https://github.com/liferay/liferay-js-themes-toolkit/pull/192 I noticed that our post-install gulp task was bombing out after creating a new layout. I went back and established that the same is happening on current "master", due to this change: https://github.com/liferay/liferay-js-themes-toolkit/pull/173

Testing the previous release (v8.0.0-rc.3) confirms that this was a regression.

The way it used to work was that the post-install would run in the directory of the just-generated layout, and the generated gulpfile in that directory pulls in "liferay-plugin-node-tasks" and runs "init".

In the buggy "master" branch, however, we were instead running "init" from "liferay-js-themes-toolkit" (AKA "liferay-themes-sdk"); that is fine for generated themes, but not for generated theme layouts.

So, the fix here is obvious: run the right gulp task.

This commit is a backport from #192 suitable for application to the current "master" branch. See also this issue which describes what I learned about "liferay-plugin-nodes-tasks" while preparing this fix, plus some next steps.

https://github.com/liferay/liferay-js-themes-toolkit/issues/193

Test plan: `yo ./packages/generator-liferay-theme/generators/layout`

### DIFF
--- a/packages/generator-liferay-theme/generators/layout/index.js
+++ b/packages/generator-liferay-theme/generators/layout/index.js
@@ -130,6 +130,23 @@ var layoutGeneratorPrototype = _.merge(liferayThemeGeneratorPrototype, {
 		}
 	},
 
+	install() {
+		var skipInstall = this.options['skip-install'];
+
+		if (!skipInstall) {
+			this.installDependencies({
+				bower: false,
+				callback() {
+					const gulp = require('gulp');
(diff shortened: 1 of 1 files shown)


wincent:
Going to merge this to reduce the number of PRs in flight; will loop back to address any feedback that comes in.

PR #195, opened 2019-03-13
Update yeoman-generator v0.18.0 to 3.2.0 (for v8 of toolkit)

wincent:
This is a "backport" from https://github.com/liferay/liferay-js-themes-toolkit/pull/192 which targets the master branch (v8 of the toolkit) instead of the one that that PR targets (v9 of the toolkit).

I had to redo it substantially, as well as upgrade the "import" generator (which doesn't even exist in the v9 branch, because that one is for import Liferay Portal 6.2 themes).

Major changes between version v0.18.0 and v3.2.0 of yeoman-generator:

- You `extend` the base `Generator` class now instead of using objects.
- This in turn means that the `_.cloneDeep` and `_.merge` stuff that we were doing now to base the layout and themelet generators off of the main "app" generator no longer works. As an alternative, I extracted out the sub-methods that were in the "priority" objects and turned theme into methods that can be called explicitly.
- Furthermore, Yeoman won't fall through to using a superclass version of a "priority" method like `configuring()` or `prompting()` so you have to call `super...()` instead.
- `this.template()` was removed, requiring us to move to `this.fs.copyTpl()` instead, which requires us to explicitly use `this.templatePath()` and `this.destinationPath()`.
- Similarly, `this.directory()` gets replaced with `this.fs.copy()`.
- `this.installDependencies()` no longer takes a callback argument, requiring me to listen for an `npmInstall:end` event explicitly.
- I got rid of the `_yosay` property because the logical analog, ES6 class fields, are still only a proposal -- detailed here: https://github.com/tc39/proposal-class-fields -- and instead just used `this.options.namespace` which produces ever-so-slightly different output (ie. "liferay-theme" instead of "Liferay Theme", "liferay-theme:layout" instead of "Liferay Layout Template" etc).
- Need to change our interpolation sigil in a `_package.json` template otherwise the quotes end up getting HTML-escaped, producing invalid JSON.

After this update, our `yarn audit` report drops from:

    140 vulnerabilities found - Packages audited: 145510
    Severity: 52 Low | 43 Moderate | 45 High

to:

    111 vulnerabilities found - Packages audited: 146665
    Severity: 42 Low | 33 Moderate | 36 High

Test plan:

```
yo ./packages/generator-liferay-theme
yo ./packages/generator-liferay-theme/generators/import
yo ./packages/generator-liferay-theme/generators/layout
yo ./packages/generator-liferay-theme/generators/themelet
```

Note that to test the "import" generator you need to provide a path to a 6.2 theme, so I used

    ./packages/generator-liferay-theme/generators/import/__tests__/__fixtures__/sdk-theme

for that. You will see an error at the end when we try to run our Gulp tasks, but this is expected, I think, and also occurs on "master":

    Error: Task './packages/generator-liferay-theme/generators/import'
    is not supported for themes with version '6.2' in this version of
    'liferay-theme-tasks'

The reason is that liferay-theme-tasks/index.js uses `argv` as the set of task names, so it calls `doctor` with a task name of "./packages/generator-liferay-theme/generators/import", which hits the "default" branch instead of the `switch` in there and complains. I'll look at addressing that in a separate commit.

Also, `yarn test`.

Related: https://github.com/liferay/liferay-js-themes-toolkit/issues/97

### DIFF
--- a/packages/generator-liferay-theme/generators/app/index.js
+++ b/packages/generator-liferay-theme/generators/app/index.js
@@ -137,9 +137,12 @@ module.exports = class extends Generator {
 		if (!skipInstall) {
 			this.on('npmInstall:end', () => {
 				const gulp = require('gulp');
-				require('liferay-theme-tasks').registerTasks({
-					gulp,
-				});
+
+				// TODO: remove in v9
+				// See: https://github.com/liferay/liferay-js-themes-toolkit/issues/196
+				process.argv = process.argv.slice(0, 2).concat(['init']);
+
+				require('liferay-theme-tasks').registerTasks({gulp});
 				gulp.start('init');
 			});
 
--- a/packages/generator-liferay-theme/generators/app/index.js
+++ b/packages/generator-liferay-theme/generators/app/index.js
@@ -1,18 +1,18 @@
 'use strict';
 
-var _ = require('lodash');
-var chalk = require('chalk');
-var Insight = require('insight');
-var minimist = require('minimist');
-var path = require('path');
-var yeoman = require('yeoman-generator');
-var yosay = require('yosay');
+const _ = require('lodash');
+const chalk = require('chalk');
+const Insight = require('insight');
+const minimist = require('minimist');
+const path = require('path');
+const Generator = require('yeoman-generator');
+const yosay = require('yosay');
 
-var lookup = require('liferay-theme-tasks/lib/lookup');
+const lookup = require('liferay-theme-tasks/lib/lookup');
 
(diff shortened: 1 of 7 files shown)


wincent:
Marking this one as a draft only because it depends on https://github.com/liferay/liferay-js-themes-toolkit/pull/194 which will have to be merged first. Otherwise this one is ready for review @julien.

wincent:
Going to merge this to reduce the number of overlapping PRs in flight; will loop back to address any feedback that comes in later.

Issue #185, opened 2019-03-06
Rename to liferay-js-themes-toolkit


wincent:
We probably don't want to rename the actual repo so as not to break bookmarks (or maybe we do, I don't know if GitHub makes it possible/easy to redirect people), but at least in terms of the documentation and internal terminology we want to rebrand this as "a themes toolkit" as opposed to an SDK because it is closer in spirit to a toolkit than a set of APIs, and it's a pattern we've been following in other places, according to what I've heard from @jbalsas.

wincent:
> I don't know if GitHub makes it possible/easy to redirect people

According to [the docs](https://help.github.com/en/articles/renaming-a-repository), redirects will take effect automatically if we rename the repo.

wincent:
Repo has been renamed now; once the in-flight PRs land, will do a follow-up pass through to make sure all the internal references are updated. (For example, in my first pass at this I change "liferay-themes-sdk" to "liferay-themes-toolkit", but the final name that we settled on is "liferay-js-themes-toolkit".)

yuchi:
Now we have to fix all our training slides :)

wincent:
Yes @yuchi, although the good news is that the old URLs will still work in the meantime (even `git clone` etc will work with the old URLs).

wincent:
Closing this one because it is basically done on both the v8 (#197) and v9 branches (68b92f3af767f6511d0).

PR #197, opened 2019-03-13
Update references to old repository name (v8)

wincent:
I already did this for v9 in:

https://github.com/liferay/liferay-js-themes-toolkit/pull/187

But as we are going to cut at least one last v8 release, we should rename it there too.

Related: https://github.com/liferay/liferay-js-themes-toolkit/issues/185


wincent:
Did an over-the-shoulder review with @julien.

PR #207, opened 2019-03-14
Fix "packageVersion is not defined" generator an "All" themelet

wincent:
This is a regression that I most likely introduced in:

https://github.com/liferay/liferay-js-themes-toolkit/pull/195

The v9 version of the same:

https://github.com/liferay/liferay-js-themes-toolkit/pull/187

doesn't have the same bug because it uses a different workaround, but I like the one used in this commit better so I am going to propagate it over there too.

Basically, we can't call `lookup` to get the `devDependencies` when making an "All" versions themelet, and we don't have to anyway, because the generated "package.json" does not have a "devDependencies" property in any case.

Test plan:

- `yarn test`
- `yo ./packages/generator-liferay-theme/generators/themelet`
- `yo ./packages/generator-liferay-theme/generators/import`
- `yo ./packages/generator-liferay-theme/generators/layout`
- `yo ./packages/generator-liferay-theme`

### DIFF
--- a/packages/generator-liferay-theme/generators/app/index.js
+++ b/packages/generator-liferay-theme/generators/app/index.js
@@ -317,19 +317,21 @@ module.exports = class extends Generator {
 		const liferayVersion = props.liferayVersion;
 
 		this.appname = props.themeId;
-		this.devDependencies = JSON.stringify(
-			lookup('devDependencies', liferayVersion),
-			null,
-			2
-		)
-			.split(/\n\s*/)
-			.join('\n\t\t')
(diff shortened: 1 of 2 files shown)


wincent:
> doesn't have the same bug because it uses a different workaround, but I like the one used in this commit better so I am going to propagate it over there too.

[Here](https://github.com/liferay/liferay-js-themes-toolkit/pull/187/commits/41c232adf2d9912a2c1b1ae45efe58342d6a3874) is the propagation over to #187.

PR #209, opened 2019-03-14
Send correct MIME type (Content-Type) header for local css files

julien:


julien:
@wincent apparently we don't need JS - for some reason with JS files, those headers are send correctly but yes this is a good idea, if was actually thinking of doing like that at first but left it as simple as it could be.

julien:
@wincent updated ;)

Issue #129, opened 2019-01-24
Layouts sub-generator should generate FTL layout template file


mwilliams2014:
Velocity layout templates are deprecated in 7.1, so the generator should generate a FTL file instead of a TPL file for 7.1 projects to reflect the move.

PR #210, opened 2019-03-14
Update template paths

julien:


julien:
@mwilliams2014 this is the fix for #129 could you please review?

I tested deploying and applying the layout in portal and it works fine but I was mostly concerned about the "velocity to freemarker translation" ... although I see similar lines in [liferay-portal](https://github.com/liferay/liferay-portal/blob/master/portal-web/docroot/layouttpl/custom/1_2_1_columns_i.ftl#L4) 

Thanks!

wincent:
Based on seeing this work it looks pretty reasonable so going to merge it and we can follow up with any amendments based on feedback from Mike.

mwilliams2014:
@julien , from what I can see in the diffs, the translation looks right for the FreeMarker syntax. I'm not sure how the generator reads it and determines how many columns/rows, etc, but as long as it follows that syntax, it's right. You can see an example in our docs as well: https://dev.liferay.com/en/develop/tutorials/-/knowledge_base/7-1/creating-layout-templates-manually#anatomy

PR #211, opened 2019-03-14
Kill dead watch_socket.js and gogo-shell code

wincent:
Dead as of: https://github.com/liferay/liferay-js-themes-toolkit/pull/190

PR #212, opened 2019-03-14
Kill dead watch_socket.js and gogo-shell code (v9)

wincent:
Same as https://github.com/liferay/liferay-js-themes-toolkit/pull/211 but cherry-picked onto the "master" (v9) branch.


wincent:
As this is basically identical to #211, going to merge it. Hopefully this is last (or one of the last) changes that will need to be applied in both places, because from now on pretty much all the action should happen on master, apart from cutting the v8.0.0-rc.4 release on the 8.x/8.x-stable branches.

PR #213, opened 2019-03-14
Fold liferay-plugin-node-tasks into liferay-theme-tasks

wincent:
Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/193

PR #215, opened 2019-03-15
Remove unused "doctor" fixtures

wincent:
We removed doctor in acc9b68515a0, so we can remove these too, I think.

Test plan: `yarn test`


wincent:
Basing this one on #213, but marking it as a draft so that it doesn't get merged first by mistake. Mostly just want to see if it passes in CI.

julien:
@wincent I had a quick look too just in case - doesn't seem like we're breaking anything here

wincent:
Thanks @julien!

PR #218, opened 2019-03-15
Clean out packages directory

wincent:
I had removed these deps packages in de2245d06570466, 78257a00ef30fd3, and dc7e88bf1c7bda but they snuck, partially, back into existence in the big merge commit (2137622d7a30) done before merging the initial batch of v9 work (#187).

Let's kill them again.


wincent:
Trivial so going to merge directly.

PR #219, opened 2019-03-15
Remove lodash from generator-liferay-theme

wincent:


wincent:
Removed this from some but not all places in the generator; I think the remaining uses save for one are all in tests.

Issue #229, opened 2019-03-18
Remove references to "classic" theme


wincent:
The related issue https://github.com/liferay/liferay-js-themes-toolkit/issues/90 discusses the problems related to "classic". We recommend that people not extend it, and in v9 of the themes toolkit we are going to remove the few internal references that remain to the classic theme.

PR #230, opened 2019-03-18
Remove references to "classic" theme

wincent:
As noted in the commit messages, as I was cleaning out references to the classic theme, I ended up cleaning up some stuff in the `lookup()` module as well. In v9, we don't really need it any more, because we only really need to deal with one version.

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/229


julien:
Just started reviewing :)

wincent:
FWIW the diff here is much easier to read on the command-line with something like `git diff -M10 -b master..wincent/classic` (`-b` to ignore whitespace changes and `-M10` to detect renamed files, because otherwise `getBaseThemeDependencies.js` looks totally new instead of just being renamed and lightly edited).

Issue #225, opened 2019-03-18
Prepare v9.0.0-alpha.0 release


wincent:
- [x] ~Update dependency versions (#171)~ (done in #232)
- [x] ~Update package versions~ (done in #226)
- [x] ~Update changelog~ (done in 917095a1cbd019d57)

wincent:
Release is out: https://github.com/liferay/liferay-js-themes-toolkit/releases/tag/v9.0.0-alpha.0

Installable using `npm install generator-liferay-theme@next` (etc).

PR #232, opened 2019-03-19
Update dependencies to point at new snapshots

wincent:
These were published just yesterday. They also have an implicit dependency on "liferay-frontend-common-css" so I added that back to the list.

Related: https://github.com/liferay/liferay-js-themes-toolkit/issues/225


wincent:
Marked this one as a draft only because it is built on top of #230, and that one has to get merged first.

wincent:
Manually tested this one and it is pretty straightforward, so going to merge it.

Issue #235, opened 2019-03-20
Gulp watch should reload the browsers at any change in any CSS file.


LuismiBarcos:
Creating a new theme I am using **gulp watch**, in particular for the CSS files. I note that if the CSS file is not at the same level of folders that the _custom.scss file, watch detect it but not reload the browser. If I want to reload the browsers I have to save the _custom.scss file.

Here are the steps to reproduce the issue:
1. Create a theme: `yo liferay-theme`
2. `gulp deploy`
3. Add a new folder
4. Inside the new folder, create a new .scss file and import it to _custom.scss
5. Start **gulp watch**
6. Make changes in the new file

In the log, we can see the following:
<img width="350" alt="imagen" src="https://user-images.githubusercontent.com/17163902/54674877-9c3dfc00-4afd-11e9-8277-3f6d93b0848a.png">

If I go to _custom.scss and press cmd+s (or ctrl+s) to save the file without make changes, I can see the following in the log:
<img width="495" alt="imagen" src="https://user-images.githubusercontent.com/17163902/54675551-16bb4b80-4aff-11e9-8d82-8ea4413b060c.png">

I think that if I make changes in any scss file, **watch should reload** the browsers.

wincent:
@LuismiBarcos: if you want to test the fix before we release v8.0.1, you can pull the fixed `watch.js` file into your `node_modules` file like so:

```
# in your theme:
curl https://gist.githubusercontent.com/wincent/4a6253c552f4b8c43b545e32b9f64ad2/raw/fadc7a1ca56d4b13009b23bdb2db3a73b4fd8b84/gistfile1.txt > node_modules/liferay-theme-tasks/tasks/watch.js 
```

wincent:
And v8.0.1 is now published to NPM (along with v9.0.0-alpha.1).

PR #237, opened 2019-03-20
Make `gulp watch` pick up changes in subdirectories (v8.x)

wincent:
I suspect the test here has been broken since at least c15b0ed1a20fe. We were creating a "rootDir" relative to "./src" that might look like "css" or "css/partials" etc, and then comparing it against "css" instead of doing a prefix match.

The smallest fix would have been to change those `=== 'css'` style checks to `match(/^css\b/)` etc, but in the end I went with something that I hope is clearer and more readable. We add a `getResourceDir()` function that returns us a string like "css", "js" etc, and we stick with the `===` comparisons. I went with the "resource" terminology because I think "root" might be confusing here (root of what? the repo? the theme? the src directory within the theme?).

Note also that the path operations we are doing here are simpler than before. Now it's basically:

    relativePath = path.relative('./src', resource);

whereas before it was:

    relativePath = path.relative(
      path.join(process.cwd(), './src'),
      path.dirname(resource)
    );

This is the v9 version of the fix. Will cherry-pick to the 8.x branch as well.

Test plan: Create a theme with two `@import`-ed SCSS partials, one of them inside a nested folder. See live-reloading work with `gulp watch` when either partial is edited.

Fixes: https://github.com/liferay/liferay-js-themes-toolkit/issues/235

### DIFF
--- a/packages/liferay-theme-tasks/tasks/watch.js
+++ b/packages/liferay-theme-tasks/tasks/watch.js
@@ -24,6 +24,23 @@ const MIME_TYPES = {
 	'.js': 'text/javacript',
 };
 
+/**
+ * Splits a path into an array of path components.
+ */
+function getPathComponents(pathString) {
+	return pathString.split(path.sep);
+}
+
+/**
+ * Give a path to a resource such as "src/css/partials/_header.scss",
+ * returns the the name of the child directory under "src/" containing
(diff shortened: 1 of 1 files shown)


julien:
Just started reviewing :)

Issue #221, opened 2019-03-18
Changelogs generated by github_changelog_generator aren't great


wincent:
There are two main issues we have with the changelogs created by `github_changelog_generator`:

1. The tool seems to be confused in the presence of multiple branches, including the changes from other branches in a given version.
2. The output is noisy (too granular) and redundant.

To illustrate these, take [the changelog generated for version 8.0.0](https://github.com/liferay/liferay-js-themes-toolkit/blob/f68c2c981f2dead4f37797750e87de5d98e9033b/CHANGELOG.md).

On the first point, it is possible that we're just using the tool wrong, but you can see changes like these in the v8.0.0 changelog, where two different PRs appear even though only one of them merged into the "8.x" branch and the other one went to "master" (which is the version 9 branch):

```
Address some yarn audit issues (for toolkit v9) #217 (wincent)
Address some yarn audit issues (for toolkit v8) (#199) #216 (wincent)
```

We had similar issues in the alloy-editor repo where we have to do v1.5.x and v2.0.x releases, and so it is not guaranteed that all of the "latest" work in the repo actually targets the "latest" version. Commits and PRs could target either branch, and it's not the case that higher version numbers always mean newer releases, or that PRs that go in before or after a particular date must belong to the next version to be published after that date.

Here are the many incantations I tried to get the right changelog in the alloy-editor repo; I may be using the tool wrong, but I couldn't get it to produce the right changes for the last v1.5.x releases, and ended up editing the file manually instead (in https://github.com/liferay/alloy-editor/commit/bc2df3ac20406a7f40d57be6e916f7bebe7b3ade and https://github.com/liferay/alloy-editor/commit/bc2df3ac20406a7f40d57be6e916f7bebe7b3ade):

```
github_changelog_generator liferay/alloy-editor -t $GITHUB_ACCESS_TOKEN
github_changelog_generator liferay/alloy-editor -t $GITHUB_ACCESS_TOKEN --base
github_changelog_generator liferay/alloy-editor -t $GITHUB_ACCESS_TOKEN --base CHANGELOG-1.x.md --since-tag v1.5.19
github_changelog_generator liferay/alloy-editor -t $GITHUB_ACCESS_TOKEN --base CHANGELOG-1.x.md --since-tag v1.5.19 --future-release v2.0.0-beta.1
[...]

PR #239, opened 2019-03-21
Create new changelog generator and use it to create 9.0.0-alpha.1 changelog

wincent:
The main goal here is to address #221 by replacing `github_changelog_generator` with a tool that works by looking at the local Git object graph only and does not hit the GitHub API in any way. As the tool has fewer features (ie. it doesn't look at labels etc), it relies on us writing good PR titles and using the GitHub "Merge Pull Request" button as opposed to pushing directly, squashing or rebasing. In exchange for making this concession, we get a simpler tool that avoids many or all of the problems described in the related ticket. We may add more features down the road, but I don't think we'd want to add many.

The other key design feature is the tool is mostly designed to work by adding a new section to the top of the CHANGELOG.md rather than regenerating the whole thing every time. This means that we can apply manual edits when the simplifying assumptions that the tool makes lead to a result that we feel should be augmented with some additions. The idea is that you can freely edit prior release notes if you want, then run the tool for each new release and stage your changes non-destructively with `git add --patch`.

At this point it can work in one of two ways:

Either, generate an updated changelog by prepending the changes from the latest release to the top of the existing file. For example, if I run:

    scripts/changelog.js --version=9.0.0-alpha.1

it produces: https://gist.github.com/wincent/b6fa344917d8b2038f8045371286325c

When you run it in this mode, you'll see something like:

     ____________________
    (_)                  `
      |                  |
      |   changelog.js   |
      |   ============   |
      |                  |
      |   Reporting      |
      |   for duty!      |
      |                  |
      |__________________|
      (_)_________________)

    Wrote ./CHANGELOG.md 31209 bytes ✨

Or, regenerate the whole thing from scratch by passing the `--regenerate` flag:

    scripts/changelog.js --version=9.0.0-alpha.1 --regenerate

which produces: https://gist.github.com/wincent/57befcfad058d80de11b450274f8e06d

When you run it in this mode, you'll see something like the following; note that the "error" you see when there are no more tags is expected because the tool works by traversing all the way back through the git history:

     ____________________
    (_)                  `
      |                  |
      |   changelog.js   |
      |   ============   |
      |                  |
      |   Reporting      |
      |   for duty!      |
      |                  |
      |__________________|
      (_)_________________)

    warning: No merges detected from v8.0.0-alpha.5 to v8.0.0-alpha.6!
    command: git describe --abbrev=0 v8.0.0-alpha.1~
    stdout:
    stderr: fatal: No tags can describe '89e1d12f81eee3d8e46bc0c0ad2114a6225bbedb'.
    Try --always, or create some tags.

    No more tags found (this is not an error) 🦄
    Wrote ./CHANGELOG.md 16885 bytes ✨

Output of `scripts/changelog.js --help`:

    scripts/changelog.js [option...]

    Options:
      --from=FROM                  [default: previous tag]
      --to=TO                      [default: HEAD]
      --help
      --outfile=FILE               [default: ./CHANGELOG.md]
      --remote-url=REPOSITORY_URL  [default: inferred]
      --regenerate                 [optional: replace entire changelog]
      --version=VERSION            [required: version being released]

As you can see, it is possible to stop looking backwards before a version specified with `--from`, or stop looking forwards at a version specified with `--to`, as well as override the remote repo URL with `--remote-url` (this one is used to generate GitHub links, and the tool can usually infer it from your Git remotes). Finally, there is an `--outfile` switch that may be useful for testing. In practice, I don't expect people to need any of the switches really; I added them out only because I couldn't resist.

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/221


julien:
Just started reviewing :)

PR #241, opened 2019-03-21
Prepare v8.0.1 release

wincent:
This one is just a draft because its purpose is to see things go green in CI. If it is green, will update 8.x-stable branch accordingly and cut a release from there.


wincent:
And we are green: https://travis-ci.org/liferay/liferay-js-themes-toolkit/builds/509527527

Will update 8.x-stable presently.

PR #242, opened 2019-03-21
Fix broken links in generated changelog

wincent:
Because I forgot to pass the "v" prefix.


wincent:
Merging due to triviality.

Issue #245, opened 2019-03-22
Should be able to pass a relative path to liferay-theme:import generator


wincent:
Noticed this while attempting to repro: https://github.com/liferay/liferay-js-themes-toolkit/issues/244

I was running the local 8.x version of the generator with:

```
yo ./packages/generator-liferay-theme/generators/import
```

And found that importing a 6.2 theme works if I supply an absolute path at the prompt (eg. `/Users/greghurrell/code/liferay-js-themes-toolkit/manzanita-theme`) but not a relative path (eg. `manzanita-theme` or `./manzanita-theme`). Relevant bits from the stack-trace:

```
AssertionError [ERR_ASSERTION]: Trying to copy from a source that does not exist: /Users/greghurrell/code/liferay-js-themes-toolkit/manzanita-theme/manzanita-theme/docroot/_diffs
    at EditionInterface.exports.copy (/Users/greghurrell/code/liferay-js-themes-toolkit/node_modules/mem-fs-editor/lib/actions/copy.js:48:3)
    at module.exports._writeThemeFiles (/Users/greghurrell/code/liferay-js-themes-toolkit/packages/generator-liferay-theme/generators/import/index.js:34:11)
```

(Note the duplicated "manzanita-theme/manzanita-theme" in the path.)

wincent:
8.0.2 is now published to NPM, so closing.

Issue #244, opened 2019-03-21
gulp build is not supported for themes with version '6.2'


john-co:
Steps to reproduce:
`npm install -g generator-liferay-theme@^8.x.x`
`npm install -g liferay-theme-tasks@^8.x.x`
`yo liferay-theme:import` on a 6.2 theme (zoe-healthcare)
`gulp build` on imported theme

/home/john-co/zoe-healthcare-theme/node_modules/liferay-theme-tasks/lib/doctor.js:84
					throw new Error(
					^

Error: Task 'build' is not supported for themes with version '6.2' in this version of 'liferay-theme-tasks'
    at assertTasksSupported (/home/john-co/zoe-healthcare-theme/node_modules/liferay-theme-tasks/lib/doctor.js:84:12)
    at doctor (/home/john-co/zoe-healthcare-theme/node_modules/liferay-theme-tasks/lib/doctor.js:32:2)
    at register (/home/john-co/zoe-healthcare-theme/node_modules/liferay-theme-tasks/index.js:62:2)
    at /home/john-co/zoe-healthcare-theme/node_modules/liferay-plugin-node-tasks/index.js:37:4
    at arrayEach (/home/john-co/zoe-healthcare-theme/node_modules/liferay-plugin-node-tasks/node_modules/lodash/lodash.js:516:11)
    at Function.forEach (/home/john-co/zoe-healthcare-theme/node_modules/liferay-plugin-node-tasks/node_modules/lodash/lodash.js:9344:14)
    at Object.module.exports.registerTasks (/home/john-co/zoe-healthcare-theme/node_modules/liferay-plugin-node-tasks/index.js:36:5)
    at Object.module.exports.registerTasks (/home/john-co/zoe-healthcare-theme/node_modules/liferay-theme-tasks/index.js:23:21)
    at Object.<anonymous> (/home/john-co/zoe-healthcare-theme/gulpfile.js:12:19)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at Liftoff.handleArguments (/home/linuxbrew/.linuxbrew/lib/node_modules/gulp/bin/gulp.js:116:3)
[...]

PR #246, opened 2019-03-22
Teach liferay-theme:import generator to work with relative paths

wincent:
Test plan: `yo ./packages/generator-liferay-theme/generators/import` and see it work with each of the following kinds of path:

- `my-theme`
- `./my-theme`
- `/Users/greghurrell/code/liferay-js-themes-toolkit/my-theme`

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/245

### DIFF
--- a/packages/generator-liferay-theme/generators/import/index.js
+++ b/packages/generator-liferay-theme/generators/import/index.js
@@ -29,7 +29,7 @@ module.exports = class extends Base {
 	}
 
 	_writeThemeFiles() {
-		this.sourceRoot(this.importTheme);
+		this.sourceRoot(this.destinationRoot());
 
 		this.fs.copy(
 			this.templatePath('docroot/_diffs'),


wincent:
cc @mwilliams2014 on this one. I noticed while trying to reproduce #244 that I couldn't pass a relative path to a theme when using the "import" generator, so I'm fixing that with this PR. However, I see that in [the docs](https://dev.liferay.com/en/develop/tutorials/-/knowledge_base/7-1/running-the-upgrade-task-for-6-2-themes), we say:

> You must provide the absolute path to your 6.2 theme’s root folder. The import task does not work for relative paths.

So, after this PR (ie. in the next release, which will be v8.0.2, of the toolkit), that will no longer be true. But I think we should probably leave the docs as-is in case somebody running an older version reads them. Does that sound right?

julien:
Just started reviewing :)

mwilliams2014:
Thanks for tagging me @wincent ! I think what we'll do is update the note in the docs to say:
> If you're running a version of the toolkit prior to v8.0.2, you must provide the absolute path to your 6.2 theme’s root folder. The import task does not work for relative paths prior to v8.0.2.

PR #249, opened 2019-03-22
Fix "gulp upgrade" task for 6.2 themes (#244)

wincent:
I had mistakenly removed the gulp-prompt dependency in bea3c7035673e80 because I thought it was unused and the old version (0.1.2) was adding noise to the `yarn audit` report.

Turns out it still is used by the 6.2 "upgrade" task, but the usage didn't show up in my search because there is no explicit `require()` call, only dynamic loading via the "gulp-load-plugins" package. (And now that I look at the list of plugins in the theme tasks package.json, I see some others that I suspect may not be used anywhere, but auditing those must be a task for a different commit.)

Here we add it back, but now a much newer version (1.2.0). The `yarn audit` output is unaffected.

Test plan: `gulp upgrade` in a 6.2 theme ([gist](https://gist.github.com/wincent/30a1cad30e6c92b9d9c05a067c019ddc)); `yarn test`.

Builds on: https://github.com/liferay/liferay-js-themes-toolkit/pull/247

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/244

### DIFF
--- a/packages/liferay-theme-tasks/package.json
+++ b/packages/liferay-theme-tasks/package.json
@@ -23,6 +23,7 @@
 		"gulp-load-plugins": "^1.5.0",
 		"gulp-plumber": "^0.6.6",
 		"gulp-postcss": "^8.0.0",
+		"gulp-prompt": "^1.2.0",
 		"gulp-rename": "^1.2.0",
 		"gulp-replace-task": "^0.11.0",
 		"gulp-sourcemaps": "1.6.0",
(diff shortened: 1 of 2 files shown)


wincent:
Marking this as a draft because #247 must go in before this one, but otherwise it is ready for review.

wincent:
FWIW: `gulp build` on the upgraded tasks works as well, which is the whole reason we started this anyway.

Issue #255, opened 2019-03-22
no theme folder is generated after running `yo liferay-theme:import`


john-co:
Cannot find any imported theme after completing `yo liferay-theme:import` command. Discovered after #244 was resolved. Followed same steps in 244.

wincent:
My understanding was that it was supposed to update the theme in-place, but perhaps that's mistaken. I'm not sure if these are the right docs to be looking at, but [this page](https://dev.liferay.com/develop/tutorials/-/knowledge_base/7-0/migrating-a-6-2-theme-to-liferay-7) says:

> The theme’s modified files (the files it modified from the base theme) are copied and reorganized in a newly created `src` directory.

My interpretation of that is that the new directory should be inside the existing theme folder. Is your understanding different @john-co?

wincent:
Digging a little deeper, I do see in you test plan doc @john-co that it says "Assert zoe-healthcare-theme folder is generated".

I went back and installed v8.0.0-rc.3 of the toolkit and found that it doesn't create a new theme folder; it just updates in-place.

```
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN manzanita-theme@1.0.0 No repository field.
npm WARN manzanita-theme@1.0.0 No license field.

added 1115 packages from 577 contributors and audited 10672 packages in 58.082s
found 55 vulnerabilities (17 low, 15 moderate, 23 high)
  run `npm audit fix` to fix them, or `npm audit` for details
events.js:173
      throw er; // Unhandled 'error' event
      ^

Error: spawn gulp ENOENT
 ```

I also installed v7.2.0 to see what it would do:

```
added 652 packages from 364 contributors and audited 6715 packages in 11.994s
found 57 vulnerabilities (18 low, 16 moderate, 23 high)
  run `npm audit fix` to fix them, or `npm audit` for details


   ╭────────────────────────────────────────────────────╮
   │                                                    │
   │   Update available 8.0.0-rc.3 → 8.0.2              │
   │   Run npm i -g generator-liferay-theme to update   │
   │                                                    │
   ╰────────────────────────────────────────────────────╯

events.js:173
      throw er; // Unhandled 'error' event
      ^

Error: spawn gulp ENOENT
```

[...]

PR #260, opened 2019-03-26
fix: "import" generator performing in-place imports

wincent:
In discussing the related ticket it became clear that the new behavior of the tool was not intended: we were performing an in-place update of an imported theme instead of creating a new theme folder in the current directory.

I added some comments to the code to make it clear what is actually happening.

Test plan:

1. With a 6.2 theme one directory above us, do `yo path/to/generator` and supply "../manzanita-theme" as a path. See the new theme get created in the current directory with name "manzanita-theme".
2. With a 7.2 theme one directory below us, do `yo path/to/generator` and supply "/absolute/path/to/manzanita" as a path. See the theme get created in the current directory with name "manzanita-theme" (note that it appended "-theme" even though I had intentionally stripped that suffix from the source directory.

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/255

### DIFF
--- a/packages/generator-liferay-theme/generators/import/index.js
+++ b/packages/generator-liferay-theme/generators/import/index.js
@@ -24,19 +24,15 @@ module.exports = class extends Base {
 		super.prompting();
 	}
 
-	_enforceFolderName() {
-		this.destinationRoot(path.resolve(this.importTheme));
-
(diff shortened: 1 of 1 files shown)


julien:
Just started reviewing :)

Issue #257, opened 2019-03-22
not able to `gulp upgrade` a 7.1 theme to 7.2


john-co:
**Steps to reproduce:**
1. create a theme using `yo liferay-theme`
2. run `gulp upgrade` on theme
3. Assert console error:
```
/home/john-co/jc-70-theme/node_modules/liferay-theme-tasks/lib/doctor.js:85
					throw new Error(
					^

Error: Task 'upgrade' is not supported for themes with version '7.1' in this version of 'liferay-theme-tasks'
    at assertTasksSupported (/home/john-co/jc-70-theme/node_modules/liferay-theme-tasks/lib/doctor.js:85:12)
    at doctor (/home/john-co/jc-70-theme/node_modules/liferay-theme-tasks/lib/doctor.js:43:2)
    at register (/home/john-co/jc-70-theme/node_modules/liferay-theme-tasks/index.js:62:2)
    at /home/john-co/jc-70-theme/node_modules/liferay-plugin-node-tasks/index.js:37:4
    at arrayEach (/home/john-co/jc-70-theme/node_modules/liferay-plugin-node-tasks/node_modules/lodash/lodash.js:516:11)
    at Function.forEach (/home/john-co/jc-70-theme/node_modules/liferay-plugin-node-tasks/node_modules/lodash/lodash.js:9344:14)
    at Object.module.exports.registerTasks (/home/john-co/jc-70-theme/node_modules/liferay-plugin-node-tasks/index.js:36:5)
    at Object.module.exports.registerTasks (/home/john-co/jc-70-theme/node_modules/liferay-theme-tasks/index.js:23:21)
    at Object.<anonymous> (/home/john-co/jc-70-theme/gulpfile.js:12:19)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at Liftoff.handleArguments (/home/linuxbrew/.linuxbrew/lib/node_modules/gulp/bin/gulp.js:116:3)
    at Liftoff.execute (/home/linuxbrew/.linuxbrew/lib/node_modules/gulp/node_modules/liftoff/index.js:203:12)
[...]

PR #261, opened 2019-03-26
feat: Provide hint to update to v9.x in order to upgrade a 7.1 theme

wincent:
Test plan:

In a 7.1 theme gulpfile, change the `require` to point at this PR (ie. `require('../packages/liferay-theme-tasks')`.

Then, run `gulp upgrade` in the 7.1 theme and see:

    Error: Task 'upgrade' is not supported for themes with
    version '7.1' in this version of 'liferay-theme-tasks' -
    to upgrade a 7.1 theme please run "npm install --save-dev
    liferay-theme-tasks@9.0.0-alpha.1" and then "gulp upgrade" again

Run `npm install --save-dev liferay-theme-tasks@9.0.0-alpha.1`. Change the `require` back to `require('liferay-theme-tasks')` so that the theme will use the code we just told the user to download, then run `gulp upgrade` again and see it work.

Note that users obviously won't have to edit their gulpfiles; it's only necessary for testing in a dev environment like this one.

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/257

### DIFF
--- a/packages/liferay-theme-tasks/lib/doctor.js
+++ b/packages/liferay-theme-tasks/lib/doctor.js
@@ -12,20 +12,28 @@ const lfrThemeConfig = require('./liferay_theme_config');
 
 /* eslint-disable quote-props */
 
-// Theme version support for upgrade tasks:
-const supportedUpgradeVersions = {
-	'6.2': true,
-	'7.0': true,
-};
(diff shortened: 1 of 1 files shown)


julien:
Just started reviewing :)

Issue #265, opened 2019-03-26
Error: It's not clear which file to import for '@import "custom"'


john-co:
Steps to reproduce:
1. `npm install -g generator-liferay-themes@v8.0.3` 
2. `yo liferay-theme:import` on 6.2 theme (zoe-health)
3. `gulp upgrade`
4. answer yes or no to `? Do you want to rename custom.css to _custom.scss?`
5. `gulp deploy`
 
Assert error thrown:
```john-co@john-co:~/zoe-healthcare-theme$ gulp upgrade
[11:16:36] Using gulpfile ~/zoe-healthcare-theme/gulpfile.js
[11:16:36] Starting 'upgrade'...
[11:16:36] Starting 'upgrade:create-backup-files'...
[11:16:36] Finished 'upgrade:create-backup-files' after 77 ms
[11:16:36] Starting 'upgrade:black-list'...
[11:16:36] Finished 'upgrade:black-list' after 4.26 ms
[11:16:36] Starting 'upgrade:replace-compass'...
[11:16:36] Finished 'upgrade:replace-compass' after 4.47 ms
[11:16:36] Starting 'upgrade:convert-bootstrap'...
Wrote file: src/css/custom.css
[11:16:36] Finished 'upgrade:convert-bootstrap' after 43 ms
[11:16:36] Starting 'upgrade:config'...
[11:16:36] Finished 'upgrade:config' after 3.09 ms
[11:16:36] Starting 'upgrade:rename-core-files'...
? Do you want to rename custom.css to _custom.scss? Yes
[11:16:50] Finished 'upgrade:rename-core-files' after 14 s
[11:16:50] Starting 'upgrade:create-css-diff'...
[11:16:50] Finished 'upgrade:create-css-diff' after 27 ms
[11:16:50] Starting 'upgrade:dependencies'...
npm WARN zoe-healthcare-theme@1.0.0 No description
npm WARN zoe-healthcare-theme@1.0.0 No repository field.
npm WARN zoe-healthcare-theme@1.0.0 No license field.

audited 8584 packages in 3.803s
found 52 vulnerabilities (17 low, 14 moderate, 21 high)
  run `npm audit fix` to fix them, or `npm audit` for details
[11:16:54] Finished 'upgrade:dependencies' after 4.4 s
[11:16:54] Starting 'upgrade:create-deprecated-mixins'...
[11:16:54] Finished 'upgrade:create-deprecated-mixins' after 7.25 ms
[...]

PR #268, opened 2019-03-28
fix: failure to rename css -> scss files during 6.2 -> 7.0 upgrade (#265)

wincent:
While trying to repro the related issue I noticed that CSS files in the theme were not getting renamed despite me answering "y" at the prompts.

The cause was a broken test for `promptResults[fileName]`, where `fileName` was something like "custom.css". This doesn't actually match the `promptResults` data structure, which looks like this:

    {
      custom: {
        css: true,
      },
      main: {
        css: false,
      },
    }

(where the `true` and `false` values reflect the answers given at the prompts.)

Test plan:

- `yarn test`
- `yo ./packages/generator-liferay-theme/generators/import`
- Provide path to a 6.2 theme, eg. "./manzanita".
- In the imported theme, edit the gulpfile to `require('../packages/liferay-theme-tasks')` so that we can test this change.
- `gulp upgrade`.
- Answer "y" and "n" at the prompts:

```
? Do you want to rename custom.css to _custom.scss? Yes
? Do you want to rename main.css to main.scss? No
```

- In temporary logging statements that I added, see the first file is renamed and then deleted:

```
rename(): { dirname: '.', basename: 'custom', extname: '.css' }
delete: [ '/Users/greghurrell/code/liferay-js-themes-toolkit/manzanita-theme/src/css/custom.css' ]
```

- In the imported theme, inspect "src/css" and see that the rename took effect.

Related: https://github.com/liferay/liferay-js-themes-toolkit/issues/265

### DIFF
--- a/packages/liferay-theme-tasks/lib/upgrade/6.2/upgrade.js
+++ b/packages/liferay-theme-tasks/lib/upgrade/6.2/upgrade.js
@@ -248,7 +248,7 @@ module.exports = function(options) {
 		const baseFile = ['aui', 'main'];
 
 		const prompts = [];
-		let srcPaths = [];
+		const srcPaths = [];
 
 		_.forEach(fs.readdirSync(path.join(CWD, DIR_SRC_CSS)), function(item) {
 			const fileName = path.basename(item, '.css');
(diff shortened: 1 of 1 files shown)


julien:
Just started reviewing :)

julien:
LGTM

Issue #271, opened 2019-04-02
Update stale info in generator-liferay-theme README


wincent:
Via @mwilliams2014, the README is:

> still saying Visual Studio and Python are required for node-gyp on Windows. This appears not be true any longer true ... it looks like only node-gyp and Python are required now.

PR #272, opened 2019-04-02
doc: remove stale references to Visual Studio (#271)

wincent:
Apparently, Visual Studio is no longer required, just node-gyp (and therefore Python). I removed the references, and also removed some of the regretful emotional wording in favor of a more neutral register.

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/271


wincent:
Travis is stalled and this is only a doc update so I am not going to wait for it.

PR #275, opened 2019-04-02
feat: automatically prefetch tags before running changelog generator

wincent:
It is too easy to produce a changelog with the wrong entries in it because one's local tags are not up to date; for example:

https://github.com/liferay/alloy-editor/pull/1219

This commit makes it harder to do that by running a `git remote update` before generation. If there is an error (say, because you are not connected to the network; unlikely, I know) or you simply want to run things more quickly, you can pass the `--no-update-tags` switch.

I also allow `--update-tags` as a switch, for symmetry, but that's the default anyway.

Test plan:

Run the generator and see this in the output:

```
Fetching remote tags: run with --no-update-tags to skip
```

Break the `git remote` command by changing it to `git remote updates` and see this instead:

```
Fetching remote tags: run with --no-update-tags to skip
command: git remote updates
stdout:
stderr: error: Unknown subcommand: updates
usage: git remote [-v | --verbose]
   or: git remote add [-t <branch>] [-m <master>] [-f] [--tags | --no-tags] [--mirror=<fetch|push>] <name> <url>
   or: git remote rename <old> <new>
   or: git remote remove <name>
   or: git remote set-head <name> (-a | --auto | -d | --delete | <branch>)
   or: git remote [-v | --verbose] show [-n] <name>
   or: git remote prune [-n | --dry-run] <name>
   or: git remote [-v | --verbose] update [-p | --prune] [(<group> | <remote>)...]
   or: git remote set-branches [--add] <name> <branch>...
   or: git remote get-url [--push] [--all] <name>
   or: git remote set-url [--push] <name> <newurl> [<oldurl>]
   or: git remote set-url --add <name> <newurl>
   or: git remote set-url --delete <name> <url>

    -v, --verbose         be verbose; must be placed before a subcommand

warning: Failed to update tags: run with --no-update-tags to skip
```

Note that this is just a warning, not an error, because we allow execution to continue.

Run with `--no-update-tags` and `--update-tags` too to confirm that it disables tag fetching (or allows it).

### DIFF
--- a/scripts/changelog.js
+++ b/scripts/changelog.js
@@ -293,6 +293,7 @@ function printUsage() {
 		'  --from=FROM                  [default: previous tag]',
 		'  --to=TO                      [default: HEAD]',
 		'  --help',
+		'  --no-update-tags             [optional: disable tag prefetching]',
 		'  --outfile=FILE               [default: ./CHANGELOG.md]',
 		'  --remote-url=REPOSITORY_URL  [default: inferred]',
 		'  --regenerate                 [optional: replace entire changelog]',
@@ -357,6 +358,7 @@ function parseArgs(args) {
 	const options = {
 		outfile: './CHANGELOG.md',
 		to: 'HEAD',
+		updateTags: true,
 	};
 
 	let match;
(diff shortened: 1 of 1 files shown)


julien:
Just started reviewing :)

PR #276, opened 2019-04-02
fix: make it harder to accidentally produce duplicate entries in changelog

wincent:
Two things:

- Make sure we autodetect `git@github.com` remotes, which means we'll be less likely to run the command twice in a row (once without and once with the `--remote` switch).
- Make the "already contains reference to $VERSION" warning into a hard error, unless running with `--force`.


julien:
Just started reviewing :)

PR #281, opened 2019-04-04
fix: broken format script in package.json

wincent:
We longer use `prettier-eslint` in this repo, so update the tasks:

- "ci" used "prettier-eslint", so make it use "format:check" (a new task) instead.
- "format" used "prettier-eslint", so make it use "prettier".

and add "format:check" as mentioned above (name chose for consistency with other repos/branches).

Test plan: Run `yarn format:check` and `yarn format` and commit the updates.


wincent:
Added you as a reviewer for this @julien just to let you know about it, but it is pretty trivial so will likely just merge it.

Issue #280, opened 2019-04-04
Update references to "DXP" in user-facing messages


wincent:
As reported by @mwilliams2014, v9 of the generator prints:

```
This version of the liferay-js-themes-toolkit (9.0.0-alpha.1)
can create a liferay-theme:app for Liferay DXP v7.2

For older versions of Liferay DXP, please use v8 of the toolkit:

    npm i -g generator-liferay-theme@^8.0.0
    yo liferay-theme:app
```

but this may give the false impression that the tool only works with DXP itself, and not the CE, which is not the case.

Mike's suggested replacement is "DXP and Portal CE".

So this issue is for updating the generator, and at the same time I'll do a sweep through looking for other references to "DXP" on both branches. Based on a quick `git grep`, I think the only thing that will need to be updated on the 8.x branch is the README.md.

PR #282, opened 2019-04-04
chore: Update references to "DXP" in user-facing messages (8.x) (#280)

wincent:
On the 8.x branch, the main place we refer to "DXP" is in the README, so I changed it to say "DXP or Portal CE". Repeating that name in the capabilities table would have been very cumbersome, so I stripped out the names entirely, making them implicit.

The CHANGELOG also has a couple of references, so I updated those too. Because we're not using `github_changelog_generator` any more, we don't have to worry about the edits getting overwritten.

Related: https://github.com/liferay/liferay-js-themes-toolkit/issues/280


wincent:
Another one I've added you as reviewer for @julien just FYI, but it's simple enough that I'll just merge it.

Issue #274, opened 2019-04-02
Enabling `autoprefixer` on a theme breaks the deploy


gcmznt:
Enabling `autoprefixer` on Minium theme breaks the deploy just after the postCSS gulp plugin

wincent:
@gcmznt: what does the breakage look like? (ie. anything printed to the console?)

gcmznt:
I had this problem running the deploy task via `gradle`, running the task via `gulp` it works.
With the gradle deploy task the gulp build just fail silently when autoprefixer is piped

wincent:
@gcmznt: I'm having a bit of trouble reproducing this, because I am not sure what version you're working with at the moment.

At least on "master" (which uses v9 of the themes toolkit), I tried turning on autoprefixer in `modules/apps/frontend-theme-fjord/frontend-theme-fjord` (I am not sure where to get this Minium theme that you mentioned).

```diff
diff --git a/modules/apps/frontend-theme-fjord/frontend-theme-fjord/package.json b/modules/apps/frontend-theme-fjord/frontend-theme-fjord/package.json
index e9b18ba4b6f6..538ec5bde76a 100644
--- a/modules/apps/frontend-theme-fjord/frontend-theme-fjord/package.json
+++ b/modules/apps/frontend-theme-fjord/frontend-theme-fjord/package.json
@@ -11,7 +11,10 @@
                "baseTheme": "styled",
                "distName": "fjord-theme",
                "rubySass": false,
-               "version": "7.1"
+               "version": "7.1",
+               "postcss": [
+                       "autoprefixer"
+               ]
        },
        "main": "package.json",
        "name": "liferay-fjord-theme",
```

When I run `gradlew clean deploy` I see the deploy finish without errors and it includes this output suggesting that autoprefixer was invoked:

```
[12:11:11] gulp-postcss: main.css
autoprefixer: /Users/greghurrell/code/portal/liferay-portal/modules/apps/frontend-theme-fjord/frontend-theme-fjord/build/_css/main.css:2568:9: You should write display: flex by final spec instead of display: box
autoprefixer: /Users/greghurrell/code/portal/liferay-portal/modules/apps/frontend-theme-fjord/frontend-theme-fjord/build/_css/main.css:2635:3: You should write display: flex by final spec instead of display: box
[...]

PR #287, opened 2019-04-08
fix: silent failures when using autoprefixer (9.x) (#274)

wincent:
When deploying a theme using this config in the package.json:

    "liferayTheme": {
            "baseTheme": "styled",
            "distName": "fjord-theme",
            "rubySass": false,
            "version": "7.1",
            "postcss": [
                    "autoprefixer"
            ]
    },

the deploy would silently fail because "gulp-plumber" would swallow the error. If I am reading this background material on plumber correctly:

- https://github.com/floatdrop/gulp-plumber
- https://github.com/gulpjs/gulp/issues/91
- https://gist.github.com/floatdrop/8269868

the intent of the plug-in is to monkey patch the Gulp pipe objects such that an error in one file won't prevent the others from being processed. We added it in this repo in commit 9200922420ee2c290c7d8f658551 (Jan 2016, "Use gulp-plumber for r2 task so that sass-parse errors don't abort build process").

In practice, its black magic is causing deploys to fail inscrutably, so we're dropping it. In the event of an error, let's fail fast instead. And note, in this concrete instance, I don't think there was even an error being thrown in the autoprefixer build at all (based on my `console.log()`-ing around to see at what point it was going off the rails; the postcss run is finishing just fine.

Note that we could also switch our old v0.6.6 version of this dependency to the latest, v1.2.1, but there's nothing in the diff that stands out as being likely to fix this issue, and I'll be much happier if we can jettison this complexity:

https://github.com/floatdrop/gulp-plumber/compare/v0.6.6...v1.2.1

Test plan: In portal, in modules/apps/frontend-theme-fjord/frontend-theme-fjord, turn on autoprefixer with the config mentioned above. Run both `gradlew clean deploy` and also (the equivalent):

```
yarn run gulp deploy \
  --css-common-path ./build_gradle/frontend-css-common \
  --styled-path ../../frontend-theme/frontend-theme-styled/src/main/resources/META-INF/resources/_styled \
  --unstyled-path ../../frontend-theme/frontend-theme-unstyled/src/main/resources/META-INF/resources/_unstyled
```

and see the build output continue past the "autoprefixer" lines to finish with:

```
[13:21:39] Finished 'deploy' after 5.27 s
✨  Done in 5.97s.
```

Likewise, see the deployment reflected in the server log:

```
1 theme for fjord-theme is available for use
```

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/274

### DIFF
--- a/packages/liferay-theme-tasks/package.json
+++ b/packages/liferay-theme-tasks/package.json
@@ -20,7 +20,6 @@
 		"gulp-liferay-r2-css": "^0.0.2",
 		"gulp-livereload": "^4.0.1",
 		"gulp-load-plugins": "^1.5.0",
-		"gulp-plumber": "^0.6.6",
 		"gulp-postcss": "^8.0.0",
 		"gulp-rename": "^1.2.0",
 		"gulp-replace-task": "^0.11.0",
--- a/packages/liferay-theme-tasks/tasks/build.js
+++ b/packages/liferay-theme-tasks/tasks/build.js
@@ -252,7 +252,6 @@ module.exports = function(options) {
 					suffix: '_rtl',
 				})
 			)
-			.pipe(plugins.plumber())
 			.pipe(r2())
 			.pipe(gulp.dest(pathBuild + '/css'));
 	});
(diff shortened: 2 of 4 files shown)


julien:
Just started reviewing :)

Issue #126, opened 2018-12-18
PostCSS plugins can't receive configuration options


jbalsas:
From https://github.com/liferay/liferay-themes-sdk/issues/46#issuecomment-447167074

The current logic to generate the necessary postcss options [only accepts a list of plugin names](https://github.com/liferay/liferay-themes-sdk/blob/master/packages/liferay-theme-tasks/tasks/build/compile-css.js#L132-L141), which makes it imposible to pass down configuration to the plugins unless they specify another mechanism such as `package.json` extensions or its own `.rc` files.

We should allow this when configuring the postcss settings. One possibility would be:

```json
{
    "browserslist": ["last 4 version"],
    ...      
    "liferayTheme": {
        "baseTheme": "styled",
        "postcss": [
            "plugin1",
            {
                "name": "autoprefixer",
                "options": {
                    "cascade": false,
                    "remove": false
                },
            },
            "plugin3"
        ],
        "screenshot": "",
        "rubySass": false,
        "templateLanguage": "ftl",
        "version": "7.1"
    },
    ...
}
```

yuchi:
I don‘t think this is a good approach to option passing.

The _Babel/Eslint/Browserify Way™_ is (as far as I remember) this:

```js
({
  plugins: [
    [ "plugin-name", { ...options } ],
    "plugin-with-no-options",
    [ "blah-blah" ],
  ]
})
```

yuchi:
But looking [at PostCSS CLI](https://github.com/postcss/postcss-cli#context) you see the more idiomatic `{ [plugin]: options }` approach — but that looks like a “runtime options building” so probably having the configuration in the package.json is simply… wrong?

jbalsas:
Hey @yuchi, that other syntax has always been a bit confusing to me, but I wouldn't mind using it that at all.

Why do you think we shouldn't have the configuration in the `package.json`? What would be the alternative? Another `.rc` file?

jbalsas:
Now that I think about it... I think there's a `postcss.config.js` that might be available already... checking that

jbalsas:
Yeah, so it looks like we should be able to use [postcss-load-config](https://www.npmjs.com/package/postcss-load-config), so you can simply write a `postcss.config.js` or a `.postcssrc.js` file to configure it.

This however, requires that we call `postcss` with no params, so we'll need to do some changes.

How does that sound?

yuchi:
[...]

PR #289, opened 2019-04-08
feat: add ability to config PostCSS plug-ins (9.x) (#126)

wincent:
Prior to this commit, we would look at the plugins named in the theme's "package.json" file under "liferayTheme" -> "postcss" and load them. "autoprefixer" was whitelisted because we bundle it with liferay-theme-tasks (ie. we would just do `require('autoprefixer')`).

For the others we would go through our `resolveDependency` helper, which actually seems to be a pointless in the context of PostCSS. I'm going to remove that special casing in the next commit, in order to make it easier to review.

With this commit, we allow for configuration by any of the mechanisms supported by postcss-load-config:

https://www.npmjs.com/package/postcss-load-config

(gulp-postcss will use postcss-load-config transparently behind the scenes).

What this means is that you can add a "postcss" property to you "package.json" like this:

```
	"postcss": {
		"plugins": {
			"autoprefixer": {
				"grid": "autoplace"
			}
		}
	},
```

or, you can create one of:

- .postcssrc
- .postcssrc.js
- .postcssrc.json
- .postcssrc.yml
- postcss.config.js

If you try to supply a PostCSS config by one of these means *and* also via "liferayTheme" -> "postcss", we throw an error.

Test plan:

- Confirm that old style of config still works by building liferay-theme-fjord with "liferayTheme" -> "postcss" -> `["autoprefixer"]`.
- Add a .postcssrc file and see this error:

    PostCSS must be configured in only one place but it was configured
    in both "liferayTheme" properties of the package.json and also in
    .postcssrc

- Remove that file and set a "postcss" config in the "package.json" instead and see this error:

    PostCSS must be configured in only one place but it was configured
    in both "liferayTheme" properties of the package.json and also in
    package.json "postcss"

- Remove the duplicate config from the "liferayTheme" config and see the build work.

I would have liked to add automated tests for this, but it is basically impractical to do anything above what we're already doing in the Jest test suite (ie. running the build and seeing the hook callbacks fired). So for now I am just going with manual QA.

See the individual commit messages in the PR for the details on the clean-up that I did along the way.

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/126

### DIFF
--- a/packages/liferay-theme-tasks/lib/util.js
+++ b/packages/liferay-theme-tasks/lib/util.js
@@ -12,7 +12,6 @@ const es = require('event-stream');
 const fs = require('fs-extra');
 const log = require('fancy-log');
 const path = require('path');
-const resolve = require('resolve');
 const tar = require('tar-fs');
 
 const CUSTOM_DEP_PATH_ENV_VARIABLE_MAP = {
@@ -137,13 +136,7 @@ function resolveDependency(dependency) {
 		return customPath;
 	}
 
-	const depsPath = process.cwd();
-
-	const dependencyPath = resolve.sync(dependency, {
-		basedir: depsPath,
-	});
-
-	return path.dirname(require.resolve(dependencyPath));
+	return path.dirname(require.resolve(dependency, {paths: [process.cwd()]}));
 }
 
 function getCustomDependencyPath(dependency) {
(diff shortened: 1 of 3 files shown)


wincent:
Not sure why Travis is not showing as green, but it is green: https://travis-ci.org/liferay/liferay-js-themes-toolkit/builds/517308819

julien:
Just started reviewing :)

Issue #291, opened 2019-04-08
gulp watch doesn't deploy changes to theme template files


mwilliams2014:
Steps to reproduce:
1. run gulp watch
2. modify CSS file, existing JS file, and theme template

CSS and JS file changes are live reloaded to the page. 
Theme template changes do not appear, even with a page reload

wincent:
Thanks @mwilliams2014. So this is definitely supposed to work:

https://github.com/liferay/liferay-js-themes-toolkit/blob/d987a86e98c5929b6f51d5a7f769d8d074e38a84/packages/liferay-theme-tasks/tasks/watch.js#L287-L293

If we detect a template change, we run those tasks; most notably the `deploy:folder` one, which I would expect to make the changes visible (if note immediately via `watch:reload`, then certainly after you page reload).

I'm going to try and repro this, but in the meantime, can you tell us if anything appears in the tomcat log when you make the change? You should see it deploying the theme.

wincent:
I've got a repro now, and an idea for a fix. Will update once I've tried it out.

wincent:
Just published [v9.0.0](https://github.com/liferay/liferay-js-themes-toolkit/releases/tag/v9.0.0) and [v8.0.7](https://github.com/liferay/liferay-js-themes-toolkit/releases/tag/v8.0.7) containing this fix.

PR #297, opened 2019-04-09
fix: failure to update after non-JS/CSS changes in watch mode (#291)

wincent:
It seems that both "deploy:file" and "deploy:folder" are horribly broken, which means that on changing a template file nothing gets deployed.

If you look at the code, `fastDeploy()` looks for a non-existent "../../temp" folder relative to the theme files, so Gulp doesn't put anything there for the portal to pick up. I considered instead getting the "appServerPath" from the "liferay-theme.json" file, which does existing, and using its "temp" directory, but it is not clear what the internal structure of it should be, nor whether it will have the effect of reloading the module in the same way that the magical "deploy" folder does. In short, I think that this might have worked once upon a time in conjunction with the right Gogo Shell commands, but now our only choice is to do a slow-but-correct full deploy. "slow" is relative here, and may only be a few seconds in the typical case, and is certainly much better than "fast and broken".

Test plan: `gulp watch`, make a change to "src/templates/portal_normal.ptl", see rebuild and deploy occur, and see it reflected in tomcat log. Now make a change to "src/css/_custom.scss" and see it do a partial rebuild, and no output in tomcat log. In both cases, see the browser window reload automatically.

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/291

### DIFF
--- a/packages/liferay-theme-tasks/tasks/watch.js
+++ b/packages/liferay-theme-tasks/tasks/watch.js
@@ -276,23 +276,7 @@ module.exports = function(options) {
 	function getBuildTaskArray(resourceDir, defaultTaskArray) {
 		let taskArray = defaultTaskArray || [];
 
-		if (resourceDir === 'WEB-INF') {
-			taskArray = [
-				'build:clean',
-				'build:src',
-				'build:web-inf',
-				'deploy:folder',
-				'watch:reload',
-			];
-		} else if (resourceDir === 'templates') {
-			taskArray = [
-				'build:src',
-				'build:themelet-src',
-				'build:themelet-js-inject',
-				'deploy:folder',
-				'watch:reload',
-			];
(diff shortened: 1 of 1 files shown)


packages/liferay-theme-tasks/tasks/watch.js
```diff
@@ -309,7 +293,7 @@ module.exports = function(options) {
 		} else if (resourceDir === 'js') {
 			taskArray = ['build:src', 'watch:reload'];
 		} else {
-			taskArray = ['deploy:file'];
+			taskArray = ['deploy', 'watch:reload'];
```
wincent: "deploy" implies "build", so that will happen automatically.

wincent:
Plenty of code can be deleted as a result of this change. I'll push that in a moment as a separate PR.

PR #298, opened 2019-04-09
feat: auto-deploy when starting `gulp watch` (#291)

wincent:
This commit started with me wanting to delete the code that is now irrelevant in the face of the previous commit, but I ended up making `gulp watch` work better along the way by auto-deploying on the first run. There is a lot of red in here, so by way of explanation:

- We delete "deploy:file" and "deploy:folder" because they were broken; we delete the associated `fastDeploy()` they rely on.
- We delete "deploy:css-files" because it used the broken `fastDeploy()` too, and thanks to the recent watch rewrite, it wasn't even needed.
- We delete the associated tests in "watch.test.js", including the sole remaining test, which describes "when changing css files", because that one doesn't actually invoke the watch functionality at all; it is sad that we don't have any automated watch tests at all, but it is still better to have no "watch" tests than tests which claimed to be "watch" tests but weren't actually testing the watch mode.
- Get rid of the `--full` and `-f` switches to the watch task and instead to a full auto-deploy every time we start to watch, by replacing "build" with "deploy" in the `getCleanTaskArray()` implementation. The purpose of that function is to put us in a clean start at the time of invoking watch, and seeing as we're doing a build anyway, there is only a small additional cost to be paid to include a deploy as well.
- In turn, get rid of the useless `defaultTaskArray` parameter which wasn't used anywhere.

Test plan: `gulp watch` and see initial build and deploy. Edit a CSS file and see it live reload in browser (proof that the `deploy:css-files` task was broken and unnecessary). Edit a template file and see full deploy take place.

Related: https://github.com/liferay/liferay-js-themes-toolkit/issues/291


wincent:
Marking this one as a draft only because it depends on https://github.com/liferay/liferay-js-themes-toolkit/pull/297 and I don't want it to get accidentally merged before that one; it's actually ready for review if you want, @julien.

Issue #310, opened 2019-04-11
CSS is (partly) messed up when viewing through proxy


wincent:
Front page viewed through "watch" proxy:

![Welcome_-_Liferay](https://user-images.githubusercontent.com/7074/55968844-1422c080-5c7d-11e9-9d22-8796d5132869.png)

Turn off this `height: 75%` property on the search button element:

```html
<button class="btn search-bar-search-button btn-unstyled" aria-label="Submit" id="fusd" type="submit">
  <!-- some svg -->
</button>
```

and everything is fine:

<img width="1928" alt="Welcome_-_Liferay_and_Welcome_-_Liferay" src="https://user-images.githubusercontent.com/7074/55969312-d96d5800-5c7d-11e9-8ddc-5a75a8173cb4.png">

Issue #309, opened 2019-04-11
Gulp Watch doesn't let you sign in on proxy port


mwilliams2014:
Steps to reproduce (tested on Chrome with custom theme):
1. run gulp watch
2. Open proxy Port:9080
3. click sign in link

Expected: portal signs in

Actual: it just hangs at `http://localhost:9080/c/portal/login?p_l_id=20129`

Nothing in the tomcat log or browser console

As soon as I sign in to port:8080 I can go back on port:9080 and navigate the portal as signed in

wincent:
Should be fixed in [v8.0.8](https://github.com/liferay/liferay-js-themes-toolkit/releases/tag/v8.0.8) and [v9.1.1](https://github.com/liferay/liferay-js-themes-toolkit/releases/tag/v9.1.1).

Issue #311, opened 2019-04-11
"watch" mode proxy chokes on gzipped content


wincent:
If the portal is running with gzip turned on, the watch proxy garbles the content.

![localhost_9080](https://user-images.githubusercontent.com/7074/55969425-0f124100-5c7e-11e9-997e-2f06b273c9c5.png)

PR #312, opened 2019-04-11
fix: broken CSS and sign-in when using "watch" proxy (#309, #310)

wincent:
Our "watch" proxy was breaking the sign-in flow, CSS, and possibly other things that we don't know about yet because it was pre-pending the "livereload" script tag where it shouldn't go, instead at the end of the HTML response.

Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/309
Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/310

### DIFF
--- a/packages/liferay-theme-tasks/tasks/watch.js
+++ b/packages/liferay-theme-tasks/tasks/watch.js
@@ -18,8 +18,11 @@ const themeUtil = require('../lib/util');
 const tinylr = require('tiny-lr');
 const portfinder = require('portfinder');
 const url = require('url');
+const util = require('util');
 
 const PASSES = Object.values(passes);
+
+const accessAsync = util.promisify(fs.access);
 
 const DEPLOYMENT_STRATEGIES = themeUtil.DEPLOYMENT_STRATEGIES;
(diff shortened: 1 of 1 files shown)


wincent:
This works, but leaving it in draft until tomorrow because I want to do some more testing. (I only tested that the CSS isn't broken, the sign-in works, and that the script tag now appears at the end; tomorrow I'll do some theme edits and verify that the "livereload" aspect still works, and I'll also need to port this to the 8.x branch — I'd also like to look at addressing #311).

julien:
Just started reviewing :)

julien:
@wincent just tested and the issues mentioned in #309 and #310 seemed fixed with this change.

I still see a difference with the "hello world" portlet's background image, but I'm not sure if we should be concerned

theme

portal

wincent:
Thanks for testing @julien. I think the images should be fixed now too by that last commit ("@wincent
fix: don't stringify proxied responses unless appending (#309, #310)").

My portal was borked for most of the morning so I couldn't test this properly before but now I have and it seems to be working great.

julien:
Just started reviewing :)

PR #315, opened 2019-04-12
fix: broken CSS and sign-in when using "watch" proxy (8.x) (#309, #310)

wincent:
This is the v8 cherry-pick of the corresponding "master"/v9.x fix (#312).

I'll attach a screenshot showing me testing this with the 7.1 community binary.


julien:
Just started reviewing :)

PR #322, opened 2019-04-15
fix: invalid "this" in deploy-live:war task (9.x)

mangei:


wincent:
Thanks @mangei. I pushed an update that fixes the formatting issues (which the CI was complaining about).

I see this issue was introduced in https://github.com/liferay/liferay-js-themes-toolkit/commit/54c57711d207a6a6612ac25adbbb3cead0e7a534#diff-b2e3a6736d0fb9f5dd8ce63552c3ea0f back in July 2018, but it had evidently gone unnoticed since then. I'm wondering, @mangei, how did you detect the issue? (and relatedly, if we are swallowing errors anywhere, which might explain why we didn't catch this bug)

wincent:
> I'm wondering, @mangei, how did you detect the issue? (and relatedly, if we are swallowing errors anywhere, which might explain why we didn't catch this bug)

Answering my own question here, I see that this code is only called by the `deploy-live:war` task, which indicates a gap in our test coverage.

Issue #266, opened 2019-03-27
Support `gulp extend` without npm search


lgdd:
When using _Artifactory_ as your npm registry (and proxy registry), `gulp extend` is not working because _Artifactory_ doesn't support npm search syntax yet. Therefore, a feature request has been opened on _JFrog_: https://www.jfrog.com/jira/browse/RTFACT-18655 (any upvote is welcome).

In the meantime, we could maybe add some support to avoid using npm search and refer directly to a package URL. This is just an idea to start a conversation around this issue.

Thanks :^)

OlivierNeu:
We actually encounter this problem in our environment. In fact, this is a stumbling block only in the build of our themes in our Concourse Delivery tools. The solution of directly referencing the url of our thongs for example would effectively solve our problem while remaining agnostic of the storage solution.

wincent:
I have a draft for this started in: https://github.com/liferay/liferay-js-themes-toolkit/pull/332

wincent:
v9.x verision of the PR is https://github.com/liferay/liferay-js-themes-toolkit/pull/333

PR #332, opened 2019-04-30
feat: teach `gulp extend` to accept a package URL (8.x) (#266)

wincent:


wincent:
At this point just a WIP because I am not 100% sure what the expected behavior should be.

With this change as just-pushed, you can provide a URL of a package on a registry like https://registry.npmjs.org/my-liferay-theme/-/my-liferay-theme-1.0.0.tgz and it will install it in a temporary directory in order to get the config from its package.json. As noted in the comments, it would be great if we could just get the metadata from the registry but `npm view` doesn't work with tarball URLs, only with package names. Likewise, it would be nice if we could install just the package and not its dependencies, but `npm` can't do that, and even though `npm pack` can download the tarball for us (yay!) it doesn't unpack it, and I don't want to add a third-party dependency in order to get untar and gunzip working in a cross-platform way, so `npm install` is the lightest-weight option.

Open questions, what would we expect the resulting theme "package.json" to look like after this change? For reference, this is what it would look like in my 7.0 test theme if I do a standard NPM search and choose to extend "my-liferay-theme":

```json
{
	"devDependencies": {
		"gulp": "3.9.1",
		"liferay-theme-deps-7.0": "8.0.9",
		"liferay-theme-tasks": "8.0.9"
	},
	"keywords": [
		"liferay-theme"
	],
	"liferayTheme": {
		"baseTheme": {
			"liferayTheme": {
				"baseTheme": "styled",
				"screenshot": "",
				"rubySass": true,
				"templateLanguage": "ftl",
				"version": "7.0"
			},
			"name": "my-liferay-theme",
			"publishConfig": {
				"tag": "7_0_x"
			},
			"version": "1.0.0"
		},
		"screenshot": "",
		"rubySass": true,
		"templateLanguage": "ftl",
		"version": "7.0"
	},
	"main": "package.json",
	"name": "themes2-theme",
	"version": "1.0.0",
	"dependencies": {
		"my-liferay-theme": "1.0.0"
	}
}
```

and this is what it looks like when I choose to extend the correspond ".tgz" URL on the registry:

```json
{
	"devDependencies": {
		"gulp": "3.9.1",
		"liferay-theme-deps-7.0": "8.0.9",
		"liferay-theme-tasks": "8.0.9"
	},
	"keywords": [
		"liferay-theme"
	],
	"liferayTheme": {
		"baseTheme": {
			"liferayTheme": {
				"baseTheme": "styled",
				"screenshot": "",
				"rubySass": true,
				"templateLanguage": "ftl",
				"version": "7.0"
			},
			"name": "my-liferay-theme",
			"publishConfig": {
				"tag": "7_0_x"
			},
			"version": "1.0.0",
			"path": "https://registry.npmjs.org/my-liferay-theme/-/my-liferay-theme-1.0.0.tgz"
		},
		"screenshot": "",
		"rubySass": true,
		"templateLanguage": "ftl",
		"version": "7.0"
	},
	"main": "package.json",
	"name": "themes2-theme",
	"version": "1.0.0",
	"dependencies": {
		"my-liferay-theme": "https://registry.npmjs.org/my-liferay-theme/-/my-liferay-theme-1.0.0.tgz"
	}
}
```

As I've implemented it, we have a "path" key that probably doesn't make any sense and should be removed. I am flying a bit blind here because I don't have a test environment with an alternative registry, and I am not sure of what edge cases I may need to take into consideration.

Thoughts? cc: @julien, @jbalsas, @lgdd, @OlivierNeu.

packages/liferay-theme-tasks/lib/theme_finder.js
```diff
@@ -22,10 +26,7 @@ function getLiferayThemeModule(name, cb) {
 			name,
 		},
 		(err, pkg) => {
-			if (
-				(pkg && !pkg.liferayTheme) ||
-				(pkg && !_.includes(pkg.keywords, 'liferay-theme'))
-			) {
```
wincent: Incidental change made along the way: previously if `pkg` was `false`-y we didn't consider that to be an error, but is seems like it should be, right?

mwilliams2014:
@wincent please let me know when this is completed. If there is a story ticket for this, please create a LPS technical documentation ticket for it as well, so we don't forget to document this update. I know it's minor, but just want to make sure we don't miss anything :-)

wincent:
@mwilliams2014: I created https://issues.liferay.com/browse/LPS-95048 for the technical documentation — will assign to you once the feature is finished and merged. I believe the story ticket is: https://issues.liferay.com/browse/BLADE-432

wincent:
Ok, I've tested this out now and I think it is getting close to what we'd need.

Here's what I did:

1. Created and published a sample "automatic-octo-barnacle-theme" theme for 7.1 from which to extend using `yo liferay-theme` ([GitHub repo](https://github.com/wincent/automatic-octo-barnacle-theme), [NPM page](https://www.npmjs.com/package/automatic-octo-barnacle-theme)).
2. Created another 7.1 theme using `yo liferay-theme`.
3. Edited the gulpfile of that theme to use the code for this PR (ie. because I created the theme inside my repo, I just changed the `require()` statement to point at `../packages/liferay-theme-tasks`).
4. Ran `node_modules/.bin/gulp extend` in the theme and gave it a URL of https://registry.npmjs.org/automatic-octo-barnacle-theme/-/automatic-octo-barnacle-theme-1.0.1.tgz.
5. Made an edit to `src/css/_custom.scss` and deployed with `node_modules/.bin/gulp deploy` to see the theme working at all.
5. Removed `src/css/_custom.scss` from my theme because I want to see the CSS from the extended theme instead.
6. Deployed again with `node_modules/.bin/gulp deploy`.
7.  Hit the portal and see the CSS from "automatic-octo-barnacle-theme" has effect.

One thing still to remedy here is that the theme doctor that runs when you run a Gulp task will complain as follows because of the presence of a "path" key in the config; that key shouldn't be there anyway, and it contains the URL of the package which doesn't make much sense — it's only in there because I'm setting `fullPath` in this PR (that was she smallest change I could make to get the task to correctly record the URL), so I will need to look into a way of doing that that doesn't have this unwanted side-effect.

```
[13:15:13] Warning: you have dependencies that are installed from local modules. These should only be used for development purposes. Do not publish this npm module with those dependencies!
```

packages/liferay-theme-tasks/lib/theme_finder.js
wincent: In testing I discovered that returning `null` here broke a test: we were using `undefined` to indicate "nothing found" and `null` to indicate "not a theme", so I put back the `pkg &&` check to preserve the existing behavior.

jbalsas:
@wincent, just registering here our offline conversation... the main doubt was: would it be enough to simply prompt for a theme name and then assume it can be discoverable through the proxy? Specifying a name sounds a bit simpler than having to locate an artifact url...

wincent:
@jbalsas: I pinged the email thread that we're on to make sure we get the relevant input here. One thing to bear in mind is that we use [the package-json package](https://www.npmjs.com/package/package-json) (try saying that out loud) to [get the theme metadata from the registry](https://github.com/liferay/liferay-js-themes-toolkit/blob/1cdfc25468c230a2bc1d23faeff0847a3738ad2e/packages/liferay-theme-tasks/lib/theme_finder.js#L169-L173) without actually downloading and installing the package. So, in order to implement the "extend by name" functionality, we'd need to be sure that the registry proxy exposes that metadata — otherwise, we'd still have to do the kind of trick that we do in this PR and download the packages into a scratch directory in order to inspect it.

wincent:
Or otherwise refactor to change the order in which operations happen during theme extension (because currently, it installs at the end... we could try installing it first and *then* running the updates that require access to the theme metadata, but then we might have to deal with rolling back in the case of failure and that might not be trivial).

yuchi:
Ping @gianbe that should have some insight for this.

rozamirandar:
@jbalsas 

I just tested this following the same steps as @wincent described [here](https://github.com/liferay/liferay-js-themes-toolkit/pull/332#issuecomment-489064419) and it worked for extending themes and themelets.

I checked the resulting files in the build folder and the visual effect after deploying the theme.

wincent:
Ok, pulling the trigger, and I'll cut releases now.

Issue #349, opened 2019-06-07
Typo in MIME type for javascript files in watch mode


sumnerwarren:
An incorrect MIME type is being returned for JS files in watch mode. You can see this running `gulp watch` and executing a curl similar to:

```
➜ curl -I http://localhost:9080/o/theme-name/js/main.js
HTTP/1.1 200 OK
Content-Type: text/javacript
Date: Fri, 07 Jun 2019 12:52:59 GMT
Connection: keep-alive
```

Note the Content-Type of **text/javacript** instead of **text/javascript**.

wincent:
Good catch. Thanks for spotting and fixing this.

PR #351, opened 2019-06-07
fix: Fix typo in JS MIME type (9.x) (#349)

sumnerwarren:
9.x version of #349 fix.

Test plan:
* Run `gulp watch`
* Run `curl -I http://localhost:9080/o/theme-name/js/main.js`
* See `Content-Type: text/javascript` in output


wincent:
Thanks for this (and doing it on both branches too!).

PR #379, opened 2019-07-23
chore: add font-awesome in upgrade process

carloslancha:


packages/liferay-theme-tasks/lib/upgrade/7.1/upgrade.js
```diff
@@ -20,6 +20,13 @@ module.exports = function(options) {
 	gulp.task('upgrade:dependencies', function(cb) {
 		lfrThemeConfig.removeDependencies(['liferay-theme-deps-7.1']);
 		lfrThemeConfig.setDependencies(devDependencies.default, true);
+		lfrThemeConfig.setDependencies(
+			{
+				'liferay-font-awesome':
+					devDependencies.optional['liferay-font-awesome'],
```
jbalsas: what does `optional` mean here?

jbalsas: and which version of `liferay-font-awesome` will we be adding?

carloslancha: We're defining devDependencies versions in `devDependencies.js`, so they can be easy to find and change if necessary.

In the last pr adding the support for font-awesome on theme creation I divided devDependencies in `default` ones, which are always added to a new theme, and the `optional` ones, which are added depending on user prompt answers.

Are you thinking in a better way?

carloslancha:
I'm ok with adding a prompt for this. @marcoscv-work what do you think?

packages/liferay-theme-tasks/lib/upgrade/7.1/upgrade.js
jbalsas: That's okay, I just wasn't aware of it 👍

packages/generator-liferay-theme/generators/app/templates/src/css/custom.css
```diff
@@ -1,6 +1,12 @@
 /* These inject tags are used for dynamically creating imports for themelet styles, you can place them where ever you like in this file. */
 
 /* inject:imports */
+<% if (fontAwesome) { %>
+@import 'liferay-font-awesome/scss/font-awesome';
+@import 'liferay-font-awesome/scss/glyphicons';
+    
+@import 'clay/base';
```
jbalsas: Is this really needed? @marcoscv-work?

jbalsas:
Hey @carloslancha, I've made some adjustments and sent https://github.com/liferay/liferay-js-themes-toolkit/pull/382. Could you please take a look so we can go over it?

Issue #380, opened 2019-07-31
New themes with FontAwesome don't include FontAwesome


jbalsas:
Creating a new theme with FontAwesome fails to import the font.

**Test 1**:
- When generating a new theme for `7.2`
- Given you answer `Y` to the question about including `FontAwesome`
- Then:
    - ✅  `"fontAwesome": true` in the `package.json` of the created theme 
    - ❌  Some code like `@import 'liferay-font-awesome/scss/font-awesome'` in `_custom.scss`

**Test 2**: 
- When you run `gulp build` on the generated theme
- Then: 
    - ✅ The `build` folder contains a `font` folder with the icon fonts which is packaged in the `war` artifact

**Test 3**: 
- When you run `gulp deploy` on the generated theme
- Then: 
    - ❌ You can see fontawesome and glyphicons icons in portal

jbalsas:
Should be fixed in https://github.com/liferay/liferay-js-themes-toolkit/pull/379

PR #382, opened 2019-07-31
feat: add font-awesome in upgrade process

jbalsas:
Closes: https://github.com/liferay/liferay-js-themes-toolkit/issues/380

### DIFF
--- a/packages/liferay-theme-tasks/lib/upgrade/7.1/upgrade.js
+++ b/packages/liferay-theme-tasks/lib/upgrade/7.1/upgrade.js
@@ -21,6 +21,19 @@ module.exports = function(options) {
 		lfrThemeConfig.removeDependencies(['liferay-theme-deps-7.1']);
 		lfrThemeConfig.setDependencies(devDependencies.default, true);
 
+		if (options.includeFontAwesome) {
+			lfrThemeConfig.setDependencies(
+				{
+					'liferay-font-awesome':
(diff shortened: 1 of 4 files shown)


packages/liferay-theme-tasks/lib/upgrade/7.1/upgrade.js
```diff
@@ -98,11 +98,14 @@ module.exports = function(options) {
...
-		);
+		var taskArray = ['upgrade:config', 'upgrade:dependencies'];
+
+		if (options.includeFontAwesome) {
+			taskArray.push('upgrade:fontAwesome');
+		}
+
+		taskArray.push(cb);
+
+		runSequence.apply(this, taskArray);
```
wincent: Seeing as you asked about this @jbalsas, I think this pattern is fine. If you *really* wanted to avoid the mutation you could do:

```javascript
const tasks = [
    'upgrade:config',
    'upgrade:dependencies',
   options.includeFontAwesome &&  'upgrade:fontAwesome',
   cb
].filter(Boolean);

runSequence(...tasks);
```

One thing though, I'll add another commit on top of this to switch the `var` to `const` (we can safely use that).

jbalsas: Go for it! 👍

wincent:
I tested this with two 7.1 themes, once answering "y" and once answering "n". Upgraded theme looks right in both cases.

PR #389, opened 2019-09-05
chore: use liferay-changelog-generator instead of scripts/changelog.js

wincent:
Keeping the old file around as a stub though because we had Markdown documents in a few other places telling people they could use "scripts/changelog.js" from liferay-js-themes-toolkit. If you try to run it you'll see a message like:

    $ scripts/changelog.js --update-tags --version=v9.0.0-alpha.1 --no-update-tags
    error: changelog.js is now liferay-changelog-generator

    To use it, try:

        npx liferay-changelog-generator --update-tags --version=v9.0.0-alpha.1 --no-update-tags

    zsh: exit 1

Note that I am using liferay-changelog-generator via `npx` rather than adding it as an explicit dependency because we pretty much always want to use the latest and that means we never have to do an explicit update.


wincent:
PR that makes liferay-changelog-generator produce output that matches Prettier's: https://github.com/liferay/liferay-npm-tools/pull/230

PR #393, opened 2019-09-23
chore: update eslint, eslint-config-liferay, prettier

wincent:
Big update for eslint-config-liferay (from v3.0.0 to v10.0.0), so there are 437 problems to fix (351 of which are autofixable).

The ESLint update is big too (v5.15.2 to v6.4.0) but that only adds one more problem, taking the total to 438.

I'm going to split the autofixable fixes and the manual fixes over the next two commits so that it is clear what they are.

The Prettier update is low key (v1.16.4 to v1.18.2).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
 	"devDependencies": {
-		"eslint": "^5.15.2",
-		"eslint-config-liferay": "^3.0.0",
+		"eslint": "^6.4.0",
+		"eslint-config-liferay": "^10.0.0",
 		"jest": "^24.1.0",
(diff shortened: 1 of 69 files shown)


wincent:
I noted some minor glitches in the import ordering here: https://github.com/liferay/eslint-config-liferay/issues/94 but I don't think that should block this.

I also don't think I'll bother applying this to 8.x until it is clear that we need to make more changes on that branch; I think we want to keep that one in "low maintenance" mode if we can.

julien:
Just started reviewing :)

Issue #415, opened 2019-12-13
Add support for 7.3 themes to toolkit v9


wincent:
As per [this Slack thread](https://liferay.slack.com/archives/C03CAD9EF/p1576201699109800).

Basically analogous to what we did to prepare for 7.2 in [#147](https://github.com/liferay/liferay-js-themes-toolkit/issues/147).

PR #420, opened 2020-01-13
feat: adapt themelet generator to 7.3 (#415)

izaera:


packages/generator-liferay-theme/generators/common/messages.js
```diff
@@ -9,15 +9,15 @@ const chalk = require('chalk');
 const {version} = require('../../package.json');
 
 function getVersionSupportMessage(generatorNamespace) {
-	const supportedVersion = chalk.red('Liferay DXP and Portal CE v7.2');
+	const supportedVersion = chalk.red('Liferay DXP and Portal CE v7.3');
```
wincent: I think we want v9 of the toolkit to work with 7.2 and above, not just 7.3, so the message should say something like that ("v7.2 or higher"). We weren't planning on having a major toolkit update for every new major version DXP (the v8 to v9 thing was intended to be a one-time break). Has something changed there?

izaera:
I'm closing this because we want to support both 7.2 and 7.3 in this same release, not maintaining one branch per major portal release.

PR #421, opened 2020-01-13
feat: enhance R2 error handling (#417)

izaera:

PR #422, opened 2020-01-14
feat: support 7.3 in upgrade (#415)

izaera:


izaera:
This goes on top of #421, so I'm closing it.

packages/liferay-r2/package.json
```diff
@@ -0,0 +1,37 @@
+{
+	"name": "liferay-r2",
+	"description": "A fork of the CSS LTR ∞ RTL converter R2 for Liferay",
+	"version": "0.4.4",
```
wincent: Have we made any changes to this project after copy/pasta-ing it? Will we need to make a new release?

packages/liferay-r2/package.json
```diff
@@ -0,0 +1,37 @@
...
+	"homepage": "https://github.com/liferay-labs/r2",
+	"author": "Dustin Diaz <dustin@dustindiaz.com>",
+	"maintainers": [
+		"Chema Balsas <jose.balsas@liferay.com>"
+	],
+	"main": "r2.js",
+	"preferGlobal": true,
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/ded/R2.git"
```
wincent: This will need an update.

packages/liferay-theme-tasks/tasks/build.js
```diff
@@ -253,8 +254,22 @@ module.exports = function(options) {
...
+			.on('end', () => {
+				if (r2.hasCssParseErrors) {
+					// eslint-disable-next-line no-console
+					console.log(
+						chalk.yellow.bold(
+							'\n' +
+								'Some CSS files had incorrect SASS or CSS code. The build will finish but\n' +
+								'the RTL CSS will not be generated correctly.\n' +
+								'\n' +
+								'Please addresse the issues in your source code to see this error go away.\n'
```
wincent: s/addresse/address/

packages/liferay-theme-tasks/lib/r2.js
```diff
@@ -0,0 +1,71 @@
...
+				console.log(
+					chalk.red.bold('[ERROR]'),
+					chalk.magenta(`.${path.sep}${err.filename}`),
+					chalk.cyan(
+						`(${err.position.end.line}:${err.position.end.column})`
+					),
+					chalk.red.bold(err.description)
+				);
+
+				task.hasCssParseErrors = true;
```
wincent: Is this a change from the upstream version of gulp-liferay-r2-css? (Would be nice to have one commit that dumps the copied file(s) in as-is, and then another one on top of that with the changes so that we can see exactly what they were.)

packages/liferay-r2/package.json
izaera: :grimacing: Yes. Unfortunately I forgot to create two commits: the copy and the modifications.

So..., we need to release a new version.

packages/liferay-theme-tasks/lib/r2.js
izaera: Same as liferay-r2 I think. I will separate the commits for both.

PR #424, opened 2020-01-20
feat: support generation of theme projects based on Liferay Classic Theme

izaera:


wincent:
@izaera: seeing as you split this up into commits for reviewability, I'll look at it commit-by-commit and send any comments as I go instead of all at once.

Looks like CI doesn't like something in the deps; might have to update the config, or change the deps:

```
error got@10.2.2: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
error Found incompatible module
```

packages/liferay-theme-tasks/package.json
```diff
@@ -26,10 +28,12 @@
 		"gulp-sass": "^3.2.0",
 		"gulp-sourcemaps": "1.6.0",
 		"gulp-storage": "^1.0.6",
+		"gulp-util": "^3.0.8",
 		"gulp-zip": "^3.2.0",
 		"http-proxy": "^1.17.0",
 		"inquirer": "^0.12.0",
 		"liferay-frontend-common-css": "^1.0.4",
+		"liferay-r2": "^0.4.4",
```
wincent: Semi-related thought: Rather than maintaining a separate package, I was thinking when you did the copy-paste of these files that it would be nice to drop the overhead and just bake the files directly into liferay-theme-tasks (seeing as they are used in exactly one place). The only reason I didn't comment then is it seems like a bit of a licensing headache because liferay-r2 contains third-party code, but maybe it is worth it... 🤔

Having said that, the third party code is *already* living in this repo, so maybe the licensing side isn't made any worse by folding the packages together (we'd still probably need to adjust the top-level license file though).

Any thoughts?

packages/generator-liferay-theme/generators/classic/classic.js
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+
+	_getLatestThemeVersion(themeVersions, liferayVersion) {
+		const themeSemverExpr = versions.theme.classic[liferayVersion];
+
+		const themeVersion = themeVersions
+			.filter(version => semver.satisfies(version, themeSemverExpr))
+			.sort((l, r) => -semver.compare(l, r))
+			.find(version => semver.satisfies(version, themeSemverExpr));
```
wincent: Do you need that `find`? The `filter` call means only satisfying versions remain in the list, so your `find` will always return the first item in the array (or `undefined`) if the array is empty. In other words, `.find(version => semver.satisfies(version, themeSemverExpr))` looks like a long way of saying `[0]`.

packages/generator-liferay-theme/generators/classic/classic.js
```diff
@@ -0,0 +1,158 @@
...
+
+			const mavenMetadata = await new xml2js.Parser().parseStringPromise(
+				body
+			);
+
+			const themeVersions =
+				mavenMetadata.metadata.versioning[0].versions[0].version;
+
+			if (themeVersions === undefined) {
+				throw new Error(
```
wincent: It seems a code smell to `throw` an exception only to `catch` it 5 lines below (see "[exceptions as flow control](https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why)").

packages/generator-liferay-theme/lib/versions.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * © 2017 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/* eslint-disable quote-props */
```
wincent: Why is this needed?

packages/generator-liferay-theme/lib/Copier.js
```diff
@@ -0,0 +1,75 @@
...
+	 * @param  {String} src path to template
+	 * @param  {Object} context optional context object to use when
+	 * 						instantiating the template (defaults to {})
+	 */
+	copyDir(src, {context = {}} = {}) {
+		const gen = this._generator;
+		const files = fs.readdirSync(gen.templatePath(src));
+
+		files.forEach(file => {
+			if (file === '.DS_Store') {
```
wincent: ❤️ 🍏

packages/generator-liferay-theme/lib/Project.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * © 2017 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/**
+ * A class to help copying Yeoman templates.
```
wincent: This comment looks wrong (copy-pasted) from the `Copier` class.

packages/generator-liferay-theme/lib/util.js
```diff
@@ -0,0 +1,111 @@
...
+const gulp = require('gulp');
+const {
+	info,
+	print,
+	success,
+	title,
+} = require('liferay-npm-build-tools-common/lib/format');
+
+const pkgJson = require('../package.json');
+const versions = require('./versions');
```
wincent: I can't find this file in this revision. Did you forget to stage it in this commit, or is GitHub lying to me?

packages/generator-liferay-theme/lib/util.js
```diff
@@ -0,0 +1,111 @@
...
+
+const FIRST_SUPPORTED_VERSION =
+	versions.supported[versions.supported.length - 1];
+const LAST_SUPPORTED_VERSION = versions.supported[0];
```
wincent: Should we sort this list to make this less fragile in the face of possible future edits?

packages/generator-liferay-theme/generators/facet-theme/templates/package.json.ejs
```diff
@@ -1,17 +1,18 @@
 {
 	"name": "<%= themeDirName %>",
-	"version": "<%= packageVersion %>",
+	"version": "1.0.0",
 	"main": "package.json",
 	"keywords": ["liferay-theme"],
 	"liferayTheme": {
 		"baseTheme": "styled",
-		"fontAwesome": <%= fontAwesome %>,
+		"fontAwesome": false,
```
wincent: What's going on here? Can we just remove this property entirely?

wincent:
Overall looks really good @izaera. The `composeWith` refactor is quite a bit better than the old inheritance.

packages/liferay-theme-tasks/package.json
izaera: I thought the same, but for some reason decided to create a package (sadly I don't remember why). But I agree that I may revisit it and try to inline it because, in the light of some other decissions we have taken, it seems the most logical thing to do.

packages/generator-liferay-theme/generators/classic/classic.js
izaera: Functional programming :man_shrugging: . I will change it :sweat_smile: .

packages/generator-liferay-theme/generators/classic/classic.js
izaera: I just wanted to reuse the print statement, but I agree it may be cleaner to print another message there and return.

packages/generator-liferay-theme/lib/versions.js
izaera: I added it because there are two keys named `7.2` and `7.3` that I wanted to quote to make clear that they are strings, not floating point numbers, but given that the conversion is implicit and the context unambiguous, I will remove the comment.

izaera: I suppose I was "thinking in Java" instead of Javascript :sweat_smile:

packages/generator-liferay-theme/lib/Copier.js
izaera: :nauseated_face: :apple:

packages/generator-liferay-theme/lib/util.js
izaera: which file? versions.js? It's the same containing the '7.2' and '7.3' keys... :thinking: 
I can see it.

packages/generator-liferay-theme/lib/util.js
izaera: Makes sense, but it would need a custom comparator. I will add a commentary in the declaration instead. Stating that the array of versions needs to be ordered from the most recent to the oldest one.

packages/generator-liferay-theme/lib/util.js
wincent: Yeah, seeing as I reviewed this commit by commit, looking at [864c9cd8a1128241c87d54f76f773a3e02288d7e](https://github.com/liferay/liferay-js-themes-toolkit/commit/864c9cd8a1128241c87d54f76f773a3e02288d7e), which adds this `require` call, if you look at that commit's tree, there is no "versions.js" in that revision:

https://github.com/liferay/liferay-js-themes-toolkit/tree/864c9cd8a1128241c87d54f76f773a3e02288d7e/packages/generator-liferay-theme/lib

wincent:
Once request: before merging this @izaera, can you edit the PR title to be descriptive of what is happening? It's the PR title that will appear in the generated changelog, so we probably don't want it to say "fix: remove deprecation in tests".

packages/generator-liferay-theme/lib/util.js
wincent: Don't we already have one? You are already using `semver.compare` elsewhere in this package.

izaera: Sadly "7.3" is not a valid semver version :-(

imagen

izaera: Will use semver.coerce

packages/generator-liferay-theme/generators/facet-theme/templates/package.json.ejs
izaera: No, we can't.

izaera: OK. It's not there where it has to be fixed. I forgot to add a piece of code in the standard theme generator :man_facepalming: .

The facet-theme subgenerator can be used to create standard themes as well as those based in classic or admin, but we only ask for fontAwesome in the standard one, so it's its responsibility to account for that change.

However I'm not sure if the classic-based theme shouldn't have fontAwesome support... I will check.

izaera: In fact, I didn't forgot the code to modify that, but after asking it seems that classic based themes may also want to use fontAwesome, so I will change the code.

packages/generator-liferay-theme/lib/util.js
izaera: :thinking:  I will reorder the commits...

packages/liferay-theme-tasks/package.json
izaera: The third party code is MIT, as ours, so it shouldn't even pose any licensing problem.

wincent: My only doubt is that we have a [top-level LICENSE.md file](https://github.com/liferay/liferay-js-themes-toolkit/blob/master/LICENSE.md) which claims copyright over the entire project. I am not a lawyer enough to know whether that is a problem, as long as we preserve the license headers down in the nested folder where we are bundling the third-party source code (note: if this is a problem here, it is not a new one, because we were doing the same when the project lived over at [the old repo](https://github.com/liferay-labs/r2)). As you know, the MIT includes the text:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

So it is possible that we were already playing "fast and loose" with this in the last repo.

izaera:
I have added some more commits on top of this PR to automate generation of projects for QA testing so I'll close it and resend with them.

Issue #432, opened 2020-01-27
Classic and admin based themes should not propagate .css files


izaera:
Currently, the generator copies all files from admin and classic themes to the `src` folder, but this includes output .css files which are not supposed to go inside `src`.

PR #433, opened 2020-01-27
fix: exclude .css files when extracting classic/admin theme (#432)

izaera:


packages/generator-liferay-theme/lib/generation/theme-based.js
```diff
@@ -14,6 +14,7 @@ const {
 	success,
 } = require('liferay-npm-build-tools-common/lib/format');
 const path = require('path');
+const rimraf = require('rimraf');
```
wincent: Adding a new dependency! 😱

100 points off for Gryffindor!

Just think, in [NodeJs v12.10.0](https://nodejs.org/en/blog/release/v12.10.0/) we'll just be able to do `fs.rmdir(dir, { recursive: true });`...

packages/generator-liferay-theme/lib/generation/theme-based.js
```diff
@@ -199,6 +203,11 @@ function mergeThumbnailPng() {
 	fs.unlinkSync(path.resolve('src/images/thumbnail.png'));
 }
 
+function removeCssFiles() {
+	rimraf.sync('**/*.css');
+	rimraf.sync('**/*.css.map');
+}
```
wincent: Are we absolutely sure that the current working directory will always, always be the theme directory when this runs? I don't think there is any doubt, but wanted to ask, just in case.

packages/generator-liferay-theme/lib/generation/theme-based.js
izaera: Yes, but... We are requiring node 10, not 12. And I don't want to delete a directory (we have `fs-extra` for that) but some files recursively inside a directory tree. 

I thought at first I could traverse the directory and delete the files, but I prefer a newer dependency that reinventing the wheel... We have used `rimraf` in other projects and, though it has its issues in older versions, it seems stable enough to use it here.

packages/generator-liferay-theme/lib/generation/theme-based.js
izaera: Yes we are. But given that it is such a sensitive thing, I will change the code to make it explicit :+1:

izaera: I have added `.src/` to the paths to make sure we don't break too much. I tried to specify the project dir but there's no safer way to get it than assuming that it is the `cwd` :man_shrugging: .

packages/generator-liferay-theme/lib/generation/theme-based.js
wincent: Ya, ya. I was just trotting out my usual anti-dependency manifesto.

Issue #103, opened 2018-10-10
Generate a thumbnail.png from a provided screenshot.png


mwilliams2014:
The screenshot.png should be in the theme's `src/images` folder. Perhaps we can generate a dummy one that the user can simply replace.
The user's screenshot.png should match the dimensions: 1080px x 864px
The generated thumbnail.png should match the dimensions: 150px x 120px

mwilliams2014:
@jbalsas , I know this was opened 6 months ago, but I'm updating the docs for 7.2 and wanted to put this back on your radar.

wincent:
Naïve question: could we just teach `generator-liferay-theme` to copy a totally generic placeholder image into the created theme? (ie. not generate anything; just provide a totally standard placeholder).

![placeholder-2000x1440-300x216](https://user-images.githubusercontent.com/7074/55859016-6971ab80-5b71-11e9-996c-42059d7208b5.png)

or:

![download](https://user-images.githubusercontent.com/7074/55859055-80b09900-5b71-11e9-87ed-a5fac5be08c8.png)

etc

jbalsas:
Yeah, the initial idea was to simply provide a basic placeholder. 

@victorvalle, what do you think? could you provide some visuals for this?

@mwilliams2014, I don't understand the relation with the Gradle Plugin at all... how do these 2 connect?

mwilliams2014:
@jbalsas The gradle plugin expects a screenshot.png and generates a thumbnail.png from that I believe. So it's the same across the board for devs, regardless of the environment they choose to develop their theme in (theme gradle plugin or theme generator), the idea would be to require them to provide a screenshot.png for either, then one of the theme tasks or something else would generate the thumbnail.png from the screenshot.png, similar to what the theme gradle plugin does.

The idea is consistency between the two approaches to theme development. There was confusion with this for a customer who went from the theme gradle plugin to the theme generator. They were expecting to create a screenshot.png. And of course when they deployed their theme with the liferay theme tasks, their thumbnail didn't show because it expected a thumbnail.png. This would fix that confusion.

Does that make sense?

victorvalle:
Hi @mwilliams2014  the new images only need to be 480x270. The change from 4:3 to 16:9 and the card maximum width force us to this value.

mwilliams2014:
Ah, ok. Thanks for letting me know @victorvalle ! I'll update the 7.2 docs to reflect this. Is this backported to 7.0 as well, or just for 7.2?

[...]

PR #434, opened 2020-01-30
feat: add final theme thumbnail placeholder (#103)

izaera:


izaera:
I'm merging this because it is just an icon, so there's nothing to review.

PR #439, opened 2020-02-17
docs: update version ranges in README

wincent:
Since the v9.5.0 release, we support themes for 7.3 as well.


wincent:
Simple enough.

PR #441, opened 2020-02-24
chore: update for compliance with current Outbound Licensing Policy

wincent:
This change is analogous to the one I just made in:

- https://github.com/liferay/eslint-config-liferay/pull/151
- https://github.com/liferay/liferay-npm-tools/pull/394

It updates the license header format to match the latest policy:

https://grow.liferay.com/excellence/Liferay+Outbound+Licensing+Policy

Did this by updating `copyright.js`, preserving the original year, running `yarn lint:fix`, and then applying manual fixes in the `bin` files (because of its use of "shebang" line, the linter adds a redundant copy of the header). I used a Vim macro to strip out the original headers. Finally, I changed the hardcoded year in the copyright.js template (2017) to a `<%= YEAR %>`, so that subsequently created files will use the current year.

Adds a copy of the license to `LICENSES/MIT.txt`.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 /**/node_modules/*
+/copyright.js
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,5 @@
 /**
- * © 2017 Liferay, Inc. <https://liferay.com>
- *
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
  * SPDX-License-Identifier: MIT
  */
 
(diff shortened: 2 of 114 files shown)


.eslintignore
```diff
@@ -1 +1,2 @@
-/**/node_modules/*
\ No newline at end of file
+/**/node_modules/*
+/copyright.js
```
wincent: Needed to prevent linter from complaining about `YEAR` in the copyright.js template.

wincent:
Just comment changes, so merging.

PR #443, opened 2020-03-02
docs: fix bad Markdown syntax

wincent:
This Markdown was extracted from the commit message in a2bd807b8031ed6, but we can fix it manually; seeing as we won't be regenerating the changelog but rather will be appending from now on, we can expect the fix to be preserved.

Just a reminder to anybody that sees this, in order to keep the changelog quality high and free from errors like this, we should be accepting the proposed merge commit generated by GitHub when we merge in PRs and avoid hand-editing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 
 ### :package: Miscellaneous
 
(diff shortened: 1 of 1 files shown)


wincent:
Trivial syntax fix in the docs, confirmed by visual inspection. `yarn ci` works, so I am just going to merge directly.

PR #450, opened 2020-03-24
chore: Add release-snapshot script

izaera:
I'm copying the release-snapshot script from JS Toolkit to be able to automate snapshot releases and make the process less error prone.

### DIFF
--- a/scripts/release-snapshot.js
+++ b/scripts/release-snapshot.js
@@ -12,7 +12,6 @@ const readlineSync = require('readline-sync');
 const {abort, success, warn} = require('./util/report');
 const {git, yarn} = require('./util/run');
 
-const EXPECTED_PREPUBLISH = 'node ../../scripts/disable-publish.js';
 const RAW_PKG_JSON = fs.readFileSync('package.json');
 const PKG_JSON = JSON.parse(RAW_PKG_JSON);
 const RELEASE_BRANCH = ['master'];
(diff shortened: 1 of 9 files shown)


izaera:
I'm merging this because it's blocking.

scripts/release-snapshot.js
```diff
@@ -0,0 +1,122 @@
...
+
+	if (!(await isStatusClean(status))) {
+		throw 'Working copy has local changes';
+	}
+
+	success('Working copy is clean');
+
+	const branch = await git('rev-parse', '--abbrev-ref', 'HEAD');
+
+	if (branch != RELEASE_BRANCH) {
```
wincent: `!=` here but `!==` below. You should use strict equality consistently, I think, and we'll be [enforcing that via lint "soon"](https://github.com/liferay/eslint-config-liferay/issues/158).

scripts/release-snapshot.js
```diff
@@ -0,0 +1,122 @@
...
+		throw `Only releases from branch ${RELEASE_BRANCH} allowed`;
+	}
+
+	success('Branch name is OK');
+
+	try {
+		await git('pull', '--ff-only', 'upstream', branch);
```
wincent: You're assuming that people have an `upstream` remote.

Just sayin'.

scripts/release-snapshot.js
izaera: I'm removing all these in favour of liferay-js-publish, so it doesn't matter any more :man_shrugging:

PR #451, opened 2020-03-24
chore: prepare major pre-release

izaera:
Use liferay-js-publish instead of custom made scripts and increase version number in package.json files


.yarnrc
```diff
@@ -0,0 +1,2 @@
+version-tag-prefix "liferay-js-themes-toolkit/v"
```
wincent: Tags in this repo use prefix `v`, not `liferay-js-themes-toolkit/v`.

.yarnrc
```diff
@@ -0,0 +1,2 @@
...
+version-git-message "chore: prepare liferay-js-themes-toolkit/v%s"
```
wincent: And this can be simplified too: `"chore: prepare v%s release"`

The format you used here is what we would employ if this repo used independent versioning, but it doesn't.

package.json
```diff
@@ -3,11 +3,10 @@
 		"chalk": "^3.0.0",
 		"eslint": "^6.5.1",
 		"eslint-config-liferay": "^13.0.0",
-		"execa": "^4.0.0",
 		"fs-extra": "^8.1.0",
 		"jest": "^24.1.0",
+		"liferay-js-publish": "^1.0.1",
```
wincent: Here's the other radical idea. Because `liferay-js-publish` is for internal use by our team alone, and we commit to never breaking the API (because there isn't really one), we don't need to make the dependency explicitly. In the other repos where we use it, we publish using `npx liferay-js-publish` (which can be in a `package.json`'s `"scripts"` field if you want, or not). That way we don't have to run around updating multiple projects and muddying up their `yarn.lock` with info about a tool that we'll never make a breaking change to.

izaera: :thinking: Maybe we should say that's the blessed way to do it in the README -> https://github.com/liferay/liferay-npm-tools/tree/master/packages/liferay-js-publish#1-install-the-package

wincent: OK. It currently says (emphasis added):

> Either:
>
> -   Install globally: `yarn global add liferay-js-publish`; or:
> -   Add to a project: `yarn add --dev liferay-js-publish`; or:
>
> **Note that you can also just use the latest without installing (ie. via `npx liferay-js-publish`). It is safe to use liferay-js-publish as an implicit dependency because we control the package, we maintain all the projects that use the package, and we won't make any breaking changes** (or at least, we won't make any undetected breaking changes).

I'll change the order to make this "blessed" option the first thing that gets mentioned.

wincent: Here it is: https://github.com/liferay/liferay-npm-tools/pull/422

izaera:
will resend

izaera:
there's no need to change project version in package.json, I think.

PR #452, opened 2020-03-25
chore: update release process

izaera:
The aim of this PR is to integrate the `liferay-js-publish` tool into the release process to make some steps towards full automation/guidance.

My plan is to use the results of this PR to first publish a pre-release of `liferay-theme-tasks` to test it in portal and, if all goes well, then publish a production release. I will take advantage of that release to automate the process a bit more if I find it possible.


CONTRIBUTING.md
```diff
@@ -86,42 +81,25 @@ git push upstream master
 
 You can now create a draft PR (proposing a merge of "master" into "stable"); **we won't actually merge this PR; we just want to see the CI pass**.
 
-### 5. Perform the merge to `stable`
+Once we've seen the CI pass above, we can **close the PR without merging it** (it was already pushed to the "master" branch) and continue.
```
wincent: I don't think this section is right any more. You shouldn't be doing the `git commit` or `git push` to `master` that's listed above any more.

package.json
```diff
@@ -21,6 +19,7 @@
 		"lint": "eslint \".*.js\" \"*.js\" \"packages/**/*.js\" \"scripts/*.js\"",
 		"lint:fix": "eslint --fix \".*.js\" \"*.js\" \"packages/**/*.js\" \"scripts/*.js\"",
 		"qa": "node scripts/qa.js",
+		"release": "yarn workspaces release",
```
wincent: Is the order that this runs in the workspaces well-defined? I don't know if they do a lexicographical sort (or if the OS or Node ends up sorting), or if it is in inode order (probably not). It is possible that this only works by luck, if it even does work.

Lexicographical order is:

- generator-liferay-theme
- liferay-theme-mixins 
- liferay-theme-tasks

But topological order is:

- liferay-theme-tasks
- generator-liferay-theme
- liferay-theme-mixins 

(Because generator depends on tasks, or at least, it did last time I looked.)

Instead of `yarn workspaces`, make a script (`scripts/release.js` or something) that visits the packages in the correct order.

packages/generator-liferay-theme/package.json
```diff
@@ -57,6 +57,8 @@
 		"strip-ansi": "^3.0.1"
 	},
 	"scripts": {
-		"release:snapshot": "node ../../scripts/release-snapshot.js"
+		"postversion": "npx liferay-js-publish",
+		"preversion": "yarn ci",
+		"release": "yarn version"
```
wincent: In the other projects where we use `liferay-js-publish`, we tell people to use `yarn version` directly. It seems like this layer of indirection doesn't add anything here but obfuscation. By making `yarn version` explicit in the release instructions, people can immediately see what the supported options are (eg. `--patch`, `--minor`, `--major`, `--new-version` etc) without having to wonder what this `release` script really does.

package.json
wincent: Just to be clear here, I am not proposing anything complicated like an actual traversal of the dependency graph. I literally just mean a script that does (the equivalent of):

```
#!/bin/sh

(cd packages/liferay-theme-tasks && yarn version "$*")
(cd packages/generator-liferay-theme && yarn version "$*")
(cd packages/liferay-theme-mixins && yarn version "$*")
```

(doesn't literally have to be a shell script, could be a Node script for Windows compat; I just wrote it this way because it is concise.)

izaera: Mmm, as it is done now, we don't need any specific release order, because the step where we update references between packages is done before publishing.

On the other hand, it's true that if something fails in between we may leave an incorrect version in npmjs.com so it's probably better to go with the script, as you say.

packages/generator-liferay-theme/package.json
izaera: Yes, I was reading [this](https://github.com/liferay/liferay-npm-tools/issues/421) now, and was thinking about the same too :+1:

CONTRIBUTING.md
izaera: @wincent So we may remove `4` totally? 

I don't think a draft PR adds any value to the process because:

1) The release starts from a fresh working copy which has been pulled from upstream/master
2) The changelog shouldn't cause any CI problem
3) The `updatePackageVersions` may break something if it has not been correctly maintained, but that shouldn't happen, especially now that we don't hold any references to version numbers in the code and everything is retrieved from `package.json` files.
4) The js-publish tools runs a `yarn ci` locally before publishing as safety net.

package.json
wincent: > Mmm, as it is done now, we don't need any specific release order, because the step where we update references between packages is done before publishing.

Are you sure? I would expect `yarn workspaces release` to do this:

1. `cd packages/generator-liferay-theme && yarn release`
  - `release` runs `yarn version`
  - `yarn version` runs `preversion` (`yarn ci`), updates version, then `postversion`
  - `postversion` runs `liferay-js-publish`
  - `liferay-js-publish` runs `yarn publish`
  - `yarn publish` (may) fail because the version of `liferay-theme-tasks` defined in the `package.json` of `generator-liferay-theme` doesn't exist in the registry yet.
2. There is no step 2, because step 1 blew up.

---

Just testing this with this bogus package:

```
{
  "name": "wincent-test-package",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "prettier": "10.0.0"
  }
}
```

It actually let me publish it, even though it is invalid.

Of course, we should still publish them in order, because anybody trying to download such a package will get blocked by a prompt, at least with `yarn`:

```
[1/4] 🔍  Resolving packages...
Couldn't find any versions for "prettier" that matches "10.0.0"
? Please choose a version of "prettier" from this list: (Use arrow keys)
```

Admittedly a narrow window in which the race condition exists, but we should still do the right thing, I think.

CONTRIBUTING.md
wincent: Yeah, I think all of `4` can go away.

package.json
izaera: Yes. We should do it right. The other setup looks fragile and it's not so complicated.

I had the same problem with the typescript build in liferay-js-toolkit. In the end `yarn workspaces` is not so useful for these things and lerna, which was smarter and called things in order, did it much better (at least for this specific case) :man_shrugging: .

Note: I'm not advocating the return of lerna :sweat_smile:

wincent:
About 73e10180ae836ed381db4ed27e12a889276801b1, that stub script was there in case we had any other people or projects still referring to the script at that path, which is where it used to live before it was extracted. I guess enough time has passed, though, so it is probably ok to remove.

scripts/release.js
```diff
@@ -21,6 +22,20 @@ if (process.argv.length < 3) {
...
+
+confirm(`
+The changelog has been generated, please review it with your favourite editor.
+
+You may change anything you don't like and when you are finished confirm the
+following question to continue.
+
+Is the changelog correct? Shall we continue`);
+
+run('git', 'add', '.');
```
wincent: Let's not do `.`; it's needlessly dangerous. Use the specific path.

izaera:
There are several files to add in that case:

1) changelog
2) the package.json of generator-liferay-js

I'm not sure what's better: if `git add .` or `git add` for each file...

1) `git add .` is more maintainable but may add garbage to the commit
2) `git add` per file may leave some files left if we change the release script and forget to add it to the list of `git add` files (quite improbable)

Reading them, I guess I prefer 2...

wincent:
> more maintainable but may add garbage to the commit

Yep, that's the problem with `git add .` — I have seen garbage added by careless use of it many times over the years.

If you want a "safe" variant that is also "maintainable", there's `git add -u`.

izaera:
I will add a check for a clean working copy at the beginning because technically we don't know which files will be modified by `yarn updatePackageVersions` in the release script so it's better to add them all (making sure we don't add any garbage which may be lying around in the working copy).

PR #454, opened 2020-03-26
chore: switch to independent releases

izaera:
We tried to use liferay-js-publish to release all projects at once but that's
not possible now.

Thus, we have decided to change to independent versioning in this repo. This
decission is made because the life cycles of each package are different so it
makes sense to use independent versioning.

Also, we don't want to revert what has been done for liferay-js-publish because
the situation is at least as good as in liferay-npm-tools and it is a step
towards full release automation.

The only things missing as of now are:

1) Automated changelog launch
2) Some safety checks to make sure that inter-project version references are
correct.

Those will be addressed in the liferay-js-publish in the future.


CONTRIBUTING.md
```diff
@@ -40,28 +40,68 @@ All methods should be documented, following [google's format](https://github.com
...
+
+Run this in the project's folder to add the modified `package.json` and `CHANGELOG.md` files.
+
+```sh
+git add -u
+```
+
+There's no need to commit it because the next step will do it.
+
+### 3. Do the publish
```
wincent: Number seems wrong.

izaera: :man_facepalming:

CONTRIBUTING.md
```diff
@@ -40,28 +40,68 @@ All methods should be documented, following [google's format](https://github.com
...
 
 We are using [liferay-js-publish](https://github.com/liferay/liferay-npm-tools/tree/master/packages/liferay-js-publish) to perform the publication to npm and manage git tags.
 
-To perform the release, run (in the project's root folder):
+To perform the release, run (in the released project's folder):
 
 ```sh
-yarn release $VERSION
+yarn version
```
wincent: Maybe mention here the different ways of invoking this could be mentioned? 99% of the time you actually want `yarn version --patch` or `yarn version --minor` (or `--major`) because they are less error-prone than manually typing a number. (There's also `--prerelease`, `--prepatch` etc but I've never used those.)

```sh
yarn version --patch # or --minor, or --major, or --new-version
```

wincent:
I really hope this works. 😬

Issue #456, opened 2020-04-01
liferay-theme-tasks 10.0.0-alpha.1+gulp 4.0.0 - 'gulp upgrade' failed with TypeError


zhangyan0701:
**Steps to reproduce:**

1. Generate a theme of 7.2 with JS Theme Toolkit v9.5.0.
2. Edit the theme's package.json and point liferay-theme-tasks to "10.0.0-alpha.1" and gulp to "^4.0.0".
3. rm -rf node_modules (just in case).
4. rm package-lock.json (just in case).
5. Run npm install again.
6. Navigate to theme's root folder.
7. Run 'gulp upgrade'.

Expected Result:
Theme is upgraded to 7.3.

Actual Result:
Upgrade failed with below error:
![gulp upgrade](https://user-images.githubusercontent.com/25576873/78134955-963e6300-7453-11ea-9199-3f31f2e77447.png)

PR #448, opened 2020-03-23
feat: #148 support for gulp 4 / node 12

izaera:
This PR contains changes needed for #148. Right now all tests (unit and QA) are working. 

I'm still testing other features of the toolkit by hand before handling it to QA and while this gets reviewed.

PR #460, opened 2020-04-02
fix: #456 gulp upgrade fails with TypeError

izaera:
This fixes an error found during QA tests.

### DIFF
--- a/packages/generator-liferay-theme/lib/util.js
+++ b/packages/generator-liferay-theme/lib/util.js
@@ -4,7 +4,6 @@
  */
 
 const gulp = require('gulp');
-const gulp4RunSequence = require('gulp4-run-sequence');
 const {
 	info,
 	print,
@@ -19,8 +18,6 @@ const versions = require('./versions');
 const FIRST_SUPPORTED_VERSION =
 	versions.supported[versions.supported.length - 1];
 const LAST_SUPPORTED_VERSION = versions.supported[0];
-
-const runSequence = gulp4RunSequence.use(gulp);
 
 /**
  * Returns a normalized "kebab case" version of the supplied string, suitable
@@ -92,7 +89,7 @@ function runGulpInit() {
 
 	liferayThemeTasks.registerTasks({gulp});
 
-	runSequence('init');
+	gulp.series('init')();
 }
 
 /**
--- a/packages/generator-liferay-theme/lib/util.js
+++ b/packages/generator-liferay-theme/lib/util.js
@@ -4,6 +4,7 @@
  */
 
 const gulp = require('gulp');
+const gulp4RunSequence = require('gulp4-run-sequence');
 const {
 	info,
 	print,
@@ -18,6 +19,8 @@ const versions = require('./versions');
 const FIRST_SUPPORTED_VERSION =
 	versions.supported[versions.supported.length - 1];
 const LAST_SUPPORTED_VERSION = versions.supported[0];
+
+const runSequence = gulp4RunSequence.use(gulp);
 
 /**
  * Returns a normalized "kebab case" version of the supplied string, suitable
@@ -89,7 +92,7 @@ function runGulpInit() {
 
 	liferayThemeTasks.registerTasks({gulp});
 
-	gulp.start('init');
+	runSequence('init');
 }
 
 /**
--- a/packages/generator-liferay-theme/package.json
+++ b/packages/generator-liferay-theme/package.json
@@ -38,6 +38,8 @@
 		"extract-zip": "^1.6.7",
 		"fs-extra": "^8.1.0",
 		"got": "^10.2.2",
+		"gulp": "^4.0.0",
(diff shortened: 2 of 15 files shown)


wincent:
Do we need to add another dependency to solve this problem?

When I [migrated AlloyEditor from Gulp v3 to v4](https://github.com/liferay/alloy-editor/commit/a5ab24e5099d41815beaa3fba8bfc71f64d11846), I said this:

> - We no longer need the "run-sequence" dependency because we can achieve the same effect by combining `gulp.parallel()` and `gulp.series()`.

Don't know whether that applies here, but dependencies have a lot of [hidden and not-so-hidden costs](https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/dev_dependencies.md)...

wincent:
(Unsurprisingly, `gulp4-run-sequence` is [using `gulp.parallel()` and `gulp.series()` under the hood](https://github.com/electric-eloquence/gulp4-run-sequence/blob/master/index.js)).

izaera:
I used it because we already have that dependency in `liferay-theme-tasks` so it's for free. But we can remove it from both places, probably, because I moved every reference to `runSequence` to the `Project` class, so it's centralized.

wincent:
> it's for free

Not true. Just because your PR doesn't *add* the package doesn't mean that we aren't carrying the ongoing debt of having it in the dependency graph. Every dependency has a cost. `gulp4-run-sequence` is no exception. An incomplete list of examples (copy-pasted from that guidelines doc):

> -   Code that may contain latent bugs and security problems that could one day manifest as real problems. Even in popular projects, they are very few friendly eyes comprehensively auditing the entire dependency graph to detect potential issues. That means that we're effectively relying on "herd security" to keep us safe — hoping that the lion eats some other animals in the herd before it gets to us, alerting us to the risk in enough time that we can patch the vulnerability — but in the digital world, where lions aren't bounded by physical constraints and attacks can be launched indiscriminately and at scale, this is cold comfort.
> -   Code that brings a maintenance burden in terms of `yarn audit` noise (even for code that isn't actually exercised in the project).
> -   Code that could present legal problems (again, how confident are we that the entire graph is adequately audited?).
> -   Code that, despite being useful, may bring a breaking change in the future that requires a large portion of the dependency graph to be updated.
> -   Code that could be redundant with similar code elsewhere in the graph, which means it brings all of the downside costs mentioned here without really justifying itself on the upside in terms of functionality that wouldn't already have been possible using the code already in the graph.

wincent:
And additionally, I would have argued against us adding it in the first place, but I missed it because it was hidden inside (large) https://github.com/liferay/liferay-js-themes-toolkit/pull/448 (+3,725 −3,860 lines).

izaera:
:thinking: In this case I prefer to always use the same technique rather than having multiple ways of executing gulp tasks in the same project. Then, we can change all of them at once as I suggested above. 

If we leave two ways of doing the same thing, people not involved in the project may be confused when deciding which one to use.

Regarding adding it in the first place, it's not that I added it because I wanted to, it's because the previous code based on gulp 3 used it and the easiest way to migrate was that. The scope of the gulp 4 adaptation was clear: just make it work, then iterate and enhance. I didn't even adhere to that because I refactored the state handling stuff a lot, but I tried to modify the least things possible.

We cannot remove all technical debt in this project at once because it is so messy and would take much longer :-(.

izaera:
However, I will get rid of it in this PR, given that now it's easy :+1: .

wincent:
> We cannot remove all technical debt in this project at once

Yeah, that wouldn't be realistic. I'm just arguing against *adding* technical debt. I am particularly concerned about dependency-creep because it is so easy to do — it seems the default solution to any problem is often just shoveling more invisible code into the `node_modules` folder, which leads us to a bad place. The costs aren't immediately visible because the code isn't committed to the repo, and the full effect of a given dependency addition is hidden in the lock file (which GitHub conceals by default).

Now, in the concrete case of `gulp4-run-sequence`, I don't want to exaggerate the costs — it has no transitive dependencies of its own and is relatively small — but precisely because it is so trivial, I don't think we should have added it in the first place. This is not a hypothetical: I already linked a PR showing a working Gulp v3-to-v4 port that moved away from `run-sequence`. Looking at [the list of dependents on `gulp4-run-sequence`](https://www.npmjs.com/package/gulp4-run-sequence?activeTab=dependents), looks plenty of people manage to migrate to Gulp v4 without needing it: there are *only 8* dependents (compare that with [*1,294 dependents* on `run-sequence`](https://www.npmjs.com/package/run-sequence?activeTab=dependents)). And the whole reason they made `gulp.series()` and `gulp.parallel()` in v4 was to obviate the need for packages like `run-sequence`...

Of course all of that doesn't mean that it has to be removed *right now*; but neither does it mean that it wasn't a mistake to add it — I'm pretty sure it was.

izaera:
Merging this as everybody is happy now :heart_eyes:

Issue #457, opened 2020-04-01
liferay-theme-tasks 10.0.0-alpha.1+gulp 4.0.0 on Windows - Running 'gulp deploy' creates a dist folder in portal/deploy


zhangyan0701:
**Steps to reproduce:**

1. Generate a theme with JS Theme Toolkit v9.5.0.
2. Edit the theme's package.json and point liferay-theme-tasks to "10.0.0-alpha.1" and gulp to "^4.0.0".
3. rm -rf node_modules (just in case).
4. rm package-lock.json (just in case).
5. Run npm install again.
6. Navigate to theme's root folder.
7. Run 'gulp deploy'.

Expected Result:
Theme war is copied to portal/deploy folder.

Actual Result:
Dist folder is also copied to portal/deploy folder, and the theme is not deployed to portal.
![dist folder in portal deploy folder](https://user-images.githubusercontent.com/25576873/78140413-8f681e00-745c-11ea-84b3-0f86dc952612.png)

PR #461, opened 2020-04-02
fix: #457 gulp deploy fails in Windows

izaera:
See #457 for a description of what this fixes.

### DIFF
--- a/packages/liferay-theme-tasks/plugin/tasks/deploy.js
+++ b/packages/liferay-theme-tasks/plugin/tasks/deploy.js
@@ -19,7 +19,7 @@ module.exports = function() {
 		const deployPath = store.get('deployPath');
 
 		const stream = gulp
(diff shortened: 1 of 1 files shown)


izaera:
I could have investigated. But I'm lazy...

izaera:
Apparently, though the fix did it, it is not a general solution. The real solution is to use unix like paths in globs.

This seems to be _by design_ in gulp, as they [use the `glob` package under the hood](https://github.com/gulpjs/gulp/issues/2294#issuecomment-467962625) and it [explicitly requests linux paths to work](https://github.com/isaacs/node-glob#windows).

I don't know yet why it worked before. It could be that:

1) gulp 3 didn't use `glob` underneath
2) themes toolkit for gulp 3 provided linux paths as globs and I somehow broke it 

Will keep investigating.

Issue #462, opened 2020-04-03
Normalize file path handling


izaera:
Trying to fix #458 I've realised that we have problems with Windows vs POSIX paths as we had long ago in the JS Toolkit.

The root cause of #458 is that `gulp` needs glob expressions (as the ones passed to `gulp.src`) to be in POSIX format. But the globs are made based on the options received by `registerTasks`, which are provided by the user. So, if the user passes paths in Windows format, we need to convert them to POSIX before making the globs.

Also, the files `liferay-theme.json` and `liferay-plugin.json` have paths inside them that may need to be normalized too.

In the JS Toolkit I decided that every path in config files should be in POSIX format (even in Windows). This is because that way projects may be multiplatform and the same source can be used for Windows and POSIX systems. Then, the internal machinery of the JS Toolkit takes care of converting from POSIX to Windows when needed.

I would like to do the same in Themes Toolkit, but it may break compatibility with already existing projects. We may argue that because we are going to release a new major version it doesn't matter, but there's another possibility that we may use, namely that of using native paths in `registerTasks` and taking care to convert them to POSIX when needed (i.e: when running on Windows).

Still, as I said I prefer to always use POSIX paths in configuration to be multiplatform.

Opinions?

wincent:
> In the JS Toolkit I decided that every path in config files should be in POSIX format (even in Windows). This is because that way projects may be multiplatform and the same source can be used for Windows and POSIX systems. Then, the internal machinery of the JS Toolkit takes care of converting from POSIX to Windows when needed.

I think this is pretty much the *only* way to do it, really. Especially in places where paths meet globs (and `\` is used to escape). So, for consistency, even in the absence of globs, I think using POSIX paths for config is the way to go.

mwilliams2014:
At least in the case of the `liferay-theme.json`, I don't think the path format is a big issue on Windows (which I test one :-) ) because the path is automatically added for you based on the `init` tasks. If you need to change it, you can just re-run the `init` task. I'm not sure about `liferay-plugin.json`.

izaera:
@mwilliams2014 `liferay-plugin.json` is the same as `liferay-theme.json` but for layouts. So, same rules should apply.

PR #464, opened 2020-04-06
refactor: #462 normalize paths

izaera:
Make sure all paths given from outside are assumed to be in POSIX format and use them accordingly.

See #462

### DIFF
--- a/packages/liferay-theme-tasks/lib/project/options.js
+++ b/packages/liferay-theme-tasks/lib/project/options.js
@@ -18,8 +18,6 @@ class Options {
 		this._normalizeFilePath('pathDist');
 		this._normalizeFilePath('pathSrc');
 		this._normalizeFilePath('rootDir');
-
-		Object.freeze(this);
 	}
 
 	get argv() {
--- a/packages/liferay-theme-tasks/lib/project/options.js
+++ b/packages/liferay-theme-tasks/lib/project/options.js
@@ -1,0 +1,112 @@
+/**
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+const {
+	default: FilePath,
+} = require('liferay-npm-build-tools-common/lib/file-path');
+const {print, warn} = require('liferay-npm-build-tools-common/lib/format');
+
+class Options {
+	constructor(project, options) {
+		this._project = project;
+		this._options = options;
+		this._windowsPathWarned = false;
+
+		this._normalizeFilePath('pathBuild');
+		this._normalizeFilePath('pathDist');
+		this._normalizeFilePath('pathSrc');
+		this._normalizeFilePath('rootDir');
+
+		Object.freeze(this);
+	}
+
+	get argv() {
+		return this._options['argv'];
+	}
+
+	/**
+	 * Name of output artifact
+	 */
+	get distName() {
+		return this._options['distName'];
+	}
+
+	/**
+	 * Build directory as POSIX path
+	 */
+	get pathBuild() {
+		return this._options['pathBuild'];
+	}
+
+	/**
+	 * Dist directory as POSIX path
+	 */
+	get pathDist() {
+		return this._options['pathDist'];
+	}
+
+	/**
+	 * Source directory as POSIX path
+	 */
+	get pathSrc() {
+		return this._options['pathSrc'];
+	}
+
+	get postcss() {
+		return this._options['postcss'];
+	}
+
+	/**
+	 * URL prefix of resources (usually '/o')
+	 */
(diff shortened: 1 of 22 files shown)


packages/liferay-theme-tasks/lib/project/options.js
```diff
@@ -0,0 +1,112 @@
+/**
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: MIT
+ */
+
+const {
+	default: FilePath,
+} = require('liferay-npm-build-tools-common/lib/file-path');
```
wincent, line 8: Is this because the TS build doesn't use [`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop)?

Would be nice to be able to just write:

```js
const FilePath = require('...');
```

Hmm, looking at it, I [see it does](https://github.com/liferay/liferay-js-toolkit/blob/4ccb844bd2c431637a67e28a3ef60353104c9dfa/tsconfig.json#L7). Weird. Any idea what's going on here?

packages/liferay-theme-tasks/lib/project/options.js
```diff
@@ -0,0 +1,112 @@
...
+		this._project = project;
+		this._options = options;
+		this._windowsPathWarned = false;
+
+		this._normalizeFilePath('pathBuild');
+		this._normalizeFilePath('pathDist');
+		this._normalizeFilePath('pathSrc');
+		this._normalizeFilePath('rootDir');
+
+		Object.freeze(this);
```
wincent, line 22: Worried about someone overwriting `_options` or `_project` etc?

packages/liferay-theme-tasks/plugin/tasks/version.js
```diff
@@ -23,11 +22,10 @@ module.exports = function() {
 			fs.readFileSync('package.json', 'utf8')
 		).version;
 
-		var pluginPackgePropertiesPath = path.join(
-			project.options.rootDir,
+		var pluginPackgePropertiesPath = project.options.rootDir.join(
```
wincent, line 25: `Packge` is misspelled.

packages/liferay-theme-tasks/lib/project/options.js
izaera: AFAIK if you want to export a default object from TS or ES6+ you have to do it yourself. The interop takes care of the opposite: importing something from ES5 and TS/ES6+ transparently.

I guess it is because any TS/ES6+ code will export your default as `default` and the rest as named fields of the `module.exports` object, but there's no way to assign anything to `module.exports` itself, which would be what is needed to be able to require it directly.

Does that make sense?

izaera: Note I'm using ES6+ as synonym for transpiled Harmony code and ES5 as raw Node Javascript code...

packages/liferay-theme-tasks/lib/project/options.js
izaera: I'm worried about certain parts of the code putting random things inside `project.options` (that happened). I guess we shouldn't do it any more and maybe all that code has already been removed, so we can get rid of the freeze to. But wanted to make sure...

izaera: If you prefer, I can review all uses of project.options and remove the freeze...

izaera:
> Well, hard to be 100% confident that we've got the asPosix/asNative coercions right everywhere, but seems the only way to find out is to do another QA pass.

Unfortunately true :-(. I had to release 2-3 versions in the JS Toolkit when I did this same change. Hopefully QA will find any problem.

izaera:
Oops, I forgot to normalize `liferay-theme.json` and `liferay-plugin.json` :-(. I will send a following PR.

packages/liferay-theme-tasks/lib/project/options.js
wincent: > I'm worried about certain parts of the code putting random things inside `project.options` (that happened).

`Object.freeze` isn't deep, so I don't think it will help you unless you apply it recursively, and it's probably not worth it:

```
$ node
> class MyClass {
... constructor() {
..... this.options = {};
..... Object.freeze(this);
..... }
... }
undefined
> m = new MyClass()
MyClass { options: {} }
> m.options.crap = true
true
```

If you really want to keep things private, copy the options when they come in, and until we move this project to TypeScript (when you can use `this.#options`) I think it's fine to assume that anybody who touches `_options` is asking for trouble.

izaera:
Going to close & reopen this to force a travis rebuild.

izaera:
Merging this. I will send a new PR normalizing liferay-theme.json and liferay-plugin.json, as I said.

PR #465, opened 2020-04-07
refactor: #462 normalize file path handling

izaera:
See https://github.com/liferay/liferay-js-themes-toolkit/issues/462 for a description of what this PR fixes.

### DIFF
--- a/packages/liferay-theme-tasks/lib/project/store.js
+++ b/packages/liferay-theme-tasks/lib/project/store.js
@@ -10,8 +10,8 @@ const {
 const path = require('path');
 
 /**
- * List of properties that must be persisted to the .json file. All others not
- * listed here are transient (in-memory only).
+ * Collection of properties that must be persisted to the .json file. All others
+ * not listed here are transient (in-memory only).
  *
  * They are indexed by type.
  */
--- a/packages/liferay-theme-tasks/lib/project/store.js
+++ b/packages/liferay-theme-tasks/lib/project/store.js
@@ -4,43 +4,257 @@
  */
 
 const fs = require('fs-extra');
+const {
+	default: FilePath,
+} = require('liferay-npm-build-tools-common/lib/file-path');
 const path = require('path');
+
+/**
+ * List of properties that must be persisted to the .json file. All others not
+ * listed here are transient (in-memory only).
+ *
+ * They are indexed by type.
+ */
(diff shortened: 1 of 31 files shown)


packages/liferay-theme-tasks/lib/project/store.js
```diff
@@ -4,43 +4,257 @@
...
+		return this._webBundleDir;
+	}
+
+	/**
+	 * @param {FilePath | string} appServerPath
+	 */
+	set appServerPath(appServerPath) {
+		if (typeof appServerPath === 'string') {
+			appServerPath = new FilePath(appServerPath);
+		}
```
wincent, line 145: An idea here. What if you taught `new FilePath` to take `FilePath | string`; then you could do the check once there instead of here a number of times.

packages/liferay-theme-tasks/lib/project/store.js
```diff
@@ -4,43 +4,257 @@
  */
 
 const fs = require('fs-extra');
+const {
+	default: FilePath,
+} = require('liferay-npm-build-tools-common/lib/file-path');
 const path = require('path');
 
+/**
+ * List of properties that must be persisted to the .json file. All others not
```
wincent, line 13: nit: not a "List"

I'd just say "Collection" which is a nice way of avoiding having to say whether something is map/object-like, or array/list-like, or set-like etc.

packages/liferay-theme-tasks/lib/project/store.js
izaera: I had the same thought, but because FilePath is in JS Toolkit and I don't want to make a new release of everything (we need to do it for every package) decided to go this way :-(.

Hopefully we will clean all this mess once we release JS Toolkit v3 and provide a shared project with common functionality that can be released alone. My plan is to rename [liferay-npm-build-tools-common](https://github.com/liferay/liferay-js-toolkit/tree/master/packages/liferay-npm-build-tools-common) to something like `js-toolkit-core` or anything similar and put all common code there (see [#526 in JS Toolkit](https://github.com/liferay/liferay-js-toolkit/issues/526)).

izaera: Or not... :roll_eyes:

wincent: Yeah, it was just an idea for the future.

Issue #455, opened 2020-03-28
Provide an option (--config config.json) to pass in the configuration


david-truong:
I would like to provide an option to pass in the configuration so we can skip the prompts similar to what is provided for liferay-js-toolkit.

This will allow us to programmatically invoke the generators for unit tests and for blade-cli without requiring a users input.

jbalsas:
Hey @david-truong, which versions of the Toolkit will need this option? Can you provide an overview of the setup that will be integrating it?

david-truong:
We will need it for 8.x and 9.x.

For blade have a project with the following package.json

8.x
```{
	"devDependencies": {
		"generator-liferay-theme": "8.1.2",
		"yo": "3.1.0"
	},
	"scripts": {
		"build-theme": "yo liferay-theme --skip-install",
		"import-theme": "yo liferay-theme:import --skip-install"
	},
	"version": "1.0.0"
}```

9.x
```
{
	"devDependencies": {
		"generator-liferay-fragments": "1.2.1",
		"generator-liferay-js": "2.18.2",
		"generator-liferay-theme": "9.5.0",
		"generator-liferay-theme-themelet": "1.2.0",
		"yo": "3.1.0"
	},
	"scripts": {
		"build-fragment": "yo liferay-fragments",
		"build-theme": "yo liferay-theme",
		"build-widget": "yo liferay-js"
	},
	"version": "1.0.0"
}```

and we run `$YO_PATH liferay-theme --config config.json`

izaera:
See how this feature is used in https://github.com/liferay/liferay-js-themes-toolkit/pull/467

izaera:
This has been tested and works, so I'm closing it.

PR #467, opened 2020-04-08
feat: #455 support --config option

izaera:
The --config option lets users pass a JSON configuration file with predefined answers to generator questions so that they can be run in batch mode.

The syntax of the JSON files is based on the one already used in the JS Toolkit. An example would be:

```json
{
        "answers": {
                "init": {
                        "appServerPath": "/opt/tomcat",
                        "deploymentStrategy": "LocalAppServer"
                },
                "layout": {
                        "layoutName": "Layout Inside Theme"
                }
        },
        "batchMode": true
}
```

Basically, you need to provide `batchMode` to make the generator run in non-interactive mode (if not, the answers are used but only as defaults for the interactive version of the generators).

Then you fill in the `answers` section specifying the target module as the first level (f.e: init, layout, theme, ...) and then the names of the question variables inside (appServerPath, layoutName, ...).

The sections and variable names are not documented in any place, but you can search for `promptWithConfig` inside the project to get a list of all possible values.


packages/generator-liferay-theme/lib/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Get the default answer value for a given prompt.
+ * @param  {string} namespace unique string identifying the calling context
+ * @param  {string} question name of property identifying the question
+ * @param  {*} defaultDefault default value is nothing is configured
+ * @return {*} the default value for the answer
+ */
+function getDefaultAnswer(namespace, question, defaultDefault = undefined) {
```
wincent, line 41: `defaultDefault` 😂

packages/generator-liferay-theme/lib/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+			value = config.answers['*'][question];
+		}
+	}
+
+	// If not found in any section return defaultDefault
+	if (value === undefined) {
+		return defaultDefault;
+	}
+
+	// If found, return the configured value
```
wincent, line 66: This comment probably not needed. (And the one on line 61 probably isn't either.)

wincent: Probably not code golf to just do:

```js
return value === undefined ? defaultDefault : value;
```

packages/generator-liferay-theme/lib/utils/config.js
```diff
@@ -0,0 +1,89 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+
+const fs = require('fs-extra');
+const os = require('os');
+const path = require('path');
+
+let config = {
+	...safeReadJsonSync(
+		path.join(os.homedir(), '.generator-liferay-theme.json')
```
wincent, line 12: Does/would anybody actually use this?

packages/generator-liferay-theme/lib/utils/promptWithConfig.js
```diff
@@ -0,0 +1,57 @@
...
+		prompt.default = config.getDefaultAnswer(
+			namespace,
+			prompt.name,
+			defaultDefault
+		);
+
+		return prompt;
+	});
+
+	// Decide wether to run in batch or interactive mode
```
wincent, line 39: wether → whether

packages/generator-liferay-theme/lib/utils/promptWithConfig.js
```diff
@@ -0,0 +1,57 @@
...
+			if (typeof val === 'function') {
+				val = val(answers);
+			}
+
+			answers[prompt.name] = val;
+
+			return answers;
+		}, {});
+	} else {
+		return await generator.prompt(prompts);
```
wincent, line 53: Remind me: I know we tell people to run with a minimum of Node v10; do we actually force that anywhere? I guess it must be ok because you already added `await` stuff in previous PRs.

scripts/qa.js
```diff
@@ -189,11 +140,40 @@ function deleteDevDependency(pkgJson, pkgName) {
...
 
-function spawn(cmd, args, options = {}) {
-	if (options.liferayVersion && argv.liferayVersion) {
+function generateSample(cwd, generatorPath, answers) {
+	const configJsonPath = path.join(
+		os.tmpdir(),
+		'generator-liferay-theme.json'
+	);
+
+	// const prjDir = path.basename(cwd);
```
wincent, line 149: Delete.

scripts/qa.js
```diff
@@ -189,11 +140,40 @@ function deleteDevDependency(pkgJson, pkgName) {
...
+
+	const configJson = {
+		answers: {
+			init: {
+				appServerPath: '/opt/tomcat',
+				// deployPath: '/opt/deploy',
+				deploymentStrategy: 'LocalAppServer',
+				// pluginName: prjDir,
+				// url: 'http://localhost:8080',
```
wincent, line 158: Seems we shouldn't have commented-out code here either, right?

wincent:
I left some comments/questions, but this all looks fine to me. 👍

packages/generator-liferay-theme/lib/utils/config.js
izaera: :man_shrugging:

packages/generator-liferay-theme/lib/utils/config.js
izaera: It is added for coherence because we also have it in the JS Toolkit.

It's especially useful for us and may be useful for people outside too (though it is not documented so I doubt they use it).

The main purpose is to have a default value for deployment directory pointing to your local Liferay installation, but I use it for some other answers too. For example, my dot file for the JS Toolkit generator is:

```json
{
        "answers": {
                "*": {
                        "liferayPresent": true,
                        "liferayDir": "/home/ivan/Liferay/CE/bundles",
                        "sampleWanted": true,
                        "useConfiguration": false,
                        "useLocalization": false
                }
        }
}
```

This way I can press Enter->Enter->Enter when running the generator without caring about the answers.

At customer places, it could be used to default the framework choice, for instance. So, if you always use React, you don't need to look for React each time you run the generator.

packages/generator-liferay-theme/lib/utils/promptWithConfig.js
izaera: There's a check [somewhere](https://github.com/liferay/liferay-js-themes-toolkit/blob/master/packages/liferay-theme-tasks/lib/checkNodeVersion.js) but it doesn't look at the version number, only at needed features. 

Should we change that and ask for an explicit minimum version number?

scripts/qa.js
izaera: :scream: End of the World :scream:

scripts/qa.js
izaera: :scream: End of the World, the sequel :scream:

izaera: It has to be removed.

packages/generator-liferay-theme/lib/utils/promptWithConfig.js
wincent: I prefer feature sniffing in general to hard-coding version numbers, but to feature sniff `await` is hard. You can do it with something like this:

```js
function supportsAsyncAwait() {
  try {
    eval('async function fn() {}');
  } catch(error) {
    if (error instanceof SyntaxError) {
      return false;
    }
  }

  return true;
}
```

but it is quite possible that the program has already died before you even get to call this function, because you ended up loading a file that used `async`/`await` before getting the check.

I don't think I'd worry about it at this point, as I think `async`/`await` landed in Node v8, and I doubt anybody is using v6.

izaera:
I'm merging and will release an alpha for QA to test.

PR #470, opened 2020-04-15
docs: mention versions > v9 in the compatibility table

wincent:
Added blank lines as well to make the different versions more obvious.

Not currently mentioning that we're going to "deprecate" v9 as there is no great urgency for people to move off it (unless they want to use Node v12), but at some point we should mention that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,13 @@ Starting with version 9 of the toolkit, in order to keep the toolkit simple, eac
 | Capability                      | Required toolkit version |
 | ------------------------------- | ------------------------ |
 | Create themes for 6.2           | v7.x                     |
+|                                 |                          |
 | Import a 6.2 theme              | v8.x                     |
 | Create themes for 7.0           | v8.x                     |
 | Create themes for 7.1           | v8.x                     |
 | Upgrade a theme for 6.2 to 7.0  | v8.x                     |
 | Upgrade a theme for 7.0 to 7.1  | v8.x                     |
-| Upgrade a theme for 7.1 to 7.2  | v9.x                     |
-| Upgrade a theme for 7.2 to 7.3  | v9.x                     |
-| Create themes for 7.2 and above | v9.x                     |
+|                                 |                          |
+| Upgrade a theme for 7.1 to 7.2  | v9.x or above            |
+| Upgrade a theme for 7.2 to 7.3  | v9.x or above            |
+| Create themes for 7.2 and above | v9.x or above            |
--- a/README.md
+++ b/README.md
@@ -37,4 +37,4 @@ Other differences between the major versions:
 Notes:
 
 -   The 7.x series of the toolkit is unlikely to receive any further development, so is effectively deprecated.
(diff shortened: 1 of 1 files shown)


wincent:
The thing with that is if they have a working setup using v9, and they don't need Node 12 support, we don't need to force them onto v10 yet, which may require them to rewrite/adapt a bunch of custom tasks. So we need to choose our language carefully. For example, you used the word "maintenance" for v9, which I think sounds quite a bit less scary than "deprecated".

Let me see if I can come up with some wording. (But it is tricky, which is why I avoided doing it in this initial try.)

README.md
```diff
@@ -14,11 +14,27 @@ Starting with version 9 of the toolkit, in order to keep the toolkit simple, eac
...
+|                                 |                          |
+| Upgrade a theme for 7.1 to 7.2  | v9.x or above            |
+| Upgrade a theme for 7.2 to 7.3  | v9.x or above            |
+| Create themes for 7.2 and above | v9.x or above            |
+
+Other differences between the major versions:
+
+| Toolkit version | Branch   | Status             |
+| --------------- | -------- | ------------------ |
+| 7.x             | -        | deprecated         |
```
wincent, line 32: Technically, there *is* [a branch for 7.0.0](https://github.com/liferay/liferay-js-themes-toolkit/tree/archive/7.0.x), but that is from the dark old ages — we don't even have a tag for the 7.0.0 release... so I prefer not to mention it.

README.md
```diff
@@ -24,3 +24,17 @@ Starting with version 9 of the toolkit, in order to keep the toolkit simple, eac
...
+| 8.x             | `8.x`    | maintenance        |
+| 9.x             | `9.x`    | maintenance        |
+| 10.x            | `master` | active development |
+
+Notes:
+
+-   The 7.x series of the toolkit is unlikely to receive any further development, so is effectively deprecated.
+-   Most active development is taking place on the `master` branch, corresponding to the 10.x series of releases, but the 8.x and 9.x series are still valid for existing themes. You may wish to continue using v8 or v9 in order to avoid the breaking changes involved in moving to v10 (specifically, moving from Gulp v3 to Gulp v4, which may require custom theme tasks to be updated).
```
izaera, line 40: Is it true that you can do the same things with v10 as with v8? :thinking: 

What I thought is that v8 must be fully supported because it targets old Liferay versions (so you cannot upgrade to v9 or v10 until you upgrade your Liferay instance).

On the other hand, v9 vs v10 is a matter of us encouraging people to move to v10 to save us work and make the world a better place by not having two branches, but always keeping in mind that we don't want to break their setups if they don't want to migrate to v10 and that's the reason why we maintain it.

The difference is subtle but the phrase `You may wish to continue using v8 or v9 in order to avoid the breaking changes involved in moving to v10` makes me think that I can move from v8 to v10 if I wish. And I think that is not true, is it?

wincent: The thing is, I didn't want to make this more verbose than it needs to be. The first table already says which version is needed for a given DXP version.

I'll see if I can delete a couple of words here to avoid giving the impression that you received.

README.md
izaera: :grimacing:

izaera:
LGTM

wincent:
Backported this to:

- 9.x: https://github.com/liferay/liferay-js-themes-toolkit/pull/476
- 8.x: https://github.com/liferay/liferay-js-themes-toolkit/pull/477

PR #476, opened 2020-04-15
docs: backport compatibility info changes to 9.x

wincent:
This is a backport of [\#470](https://github.com/liferay/liferay-js-themes-toolkit/pull/470).

Idea is to make clear when/why to use v7, v8, v9 or v10. Backporting it here in case somebody gets onto this branch via a tag or some release notes.


izaera:
lgtm

Issue #473, opened 2020-04-15
Clean up dist tags


wincent:
With the release of 10.0.0 ([liferay-theme-tasks](https://github.com/liferay/liferay-js-themes-toolkit/releases/tag/liferay-theme-tasks%2Fv10.0.0), [generator-liferay-theme](https://github.com/liferay/liferay-js-themes-toolkit/releases/tag/generator-liferay-theme%2Fv10.0.0)), and the creation of [the `9.x` maintenance branch](https://github.com/liferay/liferay-js-themes-toolkit/tree/9.x), I tried to create a "9.x" tag to match the "8.x" tag that we already have for the v8.x series.

Turns out, NPM has tightened up their checks and you can no longer do this (you get `Tag name must not be a valid SemVer range: 9.x`). I had forgotten, if I ever knew, that NPM treats strings like "9.x" as [valid range specifiers](https://docs.npmjs.com/misc/semver) (ie. "9.x" is shorthand for ">= 9.0.0 < 10.0.0".

As such, not only do we not need `9.x` as a tag; we don't need `8.x` either. A command like `yarn add generator-liferay-theme@8.x` will behave basically the same with or without the tag. So, as step one, let's just delete the tags.

You might think that, if anything, this is better than having tags because its one less thing to do when you release a package on a maintenance branch like 8.x or 9.x. However, it is not all roses: if you cut 10.0.0, as we just did, and then cut 9.5.1 (as we are about to do), then the `latest` tag, which NPM maintains by default, will actually point at 9.5.1, which may be a bit misleading. Sure, it is literally the last-published version, but it is not the "most current" code in terms of accrued development.

As such, we *may* still need tags for those old branches. That would work ok in this repo because we are not using `liferay-js-publish` on the 8.x and 9.x branches. Tagging (or not) on those branches is still something you can explicitly do, because running `yarn publish` [is a manual process](https://github.com/liferay/liferay-js-themes-toolkit/blob/50c717b5a4d76168ef45ae1363a85c0cb16034da/CONTRIBUTING.md).

These are the packages with "excess' tags:

- https://www.npmjs.com/package/liferay-theme-tasks?activeTab=versions
- https://www.npmjs.com/package/generator-liferay-theme?activeTab=versions
- https://www.npmjs.com/package/liferay-theme-mixins?activeTab=versions

To remove them:

```
npm dist-tag rm liferay-theme-tasks 8.x
npm dist-tag rm liferay-theme-tasks next
[...]

PR #479, opened 2020-04-15
docs: update CONTRIBUTING.md

wincent:
Hey @izaera, figure we might as well make these improvements while how all this works is still fresh in our minds. Let me know what you think, especially with respect to use a generic "-pre" prefix for any prerelease.

---

Given our experience today cutting releases of v10 and v9.5.1, we spent some time looking at the effects of `.yarnrc` files, dist tags, and the behavior of `yarn version` with various different flags:

https://github.com/liferay/liferay-js-themes-toolkit/issues/473

Armed with that knowledge, I think we should make some updates to the documentation here:

- Make it clear that `yarn changelog` must run from a *package* directory, not from the top level ("project" is ambiguous; is it the repo, or a package?).
- Drop "project-name" from version number because liferay-changelog-generator will grab it from .yarnrc directly, since [this PR](https://github.com/liferay/liferay-npm-tools/pull/430).
- Provide details on version number format for prereleases.
- Switch a `git add -u` for a `git add -p`, because it is safer and provides another chance to catch mistakes.
- Document use of `--prepatch` and friends: I suggest we make this the "blessed" way to cut prereleases to avoid possible human errors.
- Fix typo ("google" -> "Google")

Note that the recommendation to use `--prepatch` etc and standardize on a prefix-id of "pre" (ie. making v1.2.3-pre.0) requires us to set a default `--preid` in a .yarnrc file. I think that always using the same "pre" id in this way makes it less likely that we'll screw up a release, and the cost of losing the distinction between "alpha", "beta" etc prereleases is a worthwhile sacrifice. We never really want customers using prereleases anyway; they are almost always used for internal testing only.

We can do this in a single root-level .yarnrc because yarn will walk up the tree looking for files, merging them. I tested this:

- When you cut a normal release (eg `yarn version --patch` etc), the `--preid` is ignored.
- When you cut a prerelease (eg. `yarn version --prepatch` etc), the `--preid` applies automatically.
- `yarn version --prerelease` does what it says in the CONTRIBUTING.md 😂

Docs: https://classic.yarnpkg.com/en/docs/cli/version

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+--version.preid pre
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ yarn format
 
 ## JS Docs
 
-All methods should be documented, following [google's format](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler).
+All methods should be documented, following [Google's format](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler).
 
 # Releasing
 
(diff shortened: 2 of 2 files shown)


izaera:
@wincent LGTM

Regarding `pre` I really don't care too much about what we use as long as it is clear that it is a pre-release and stays out of the `latest` dist-tag in npmjs.com (which is the case). 

Differences between alpha, beta, etc. are subtle and we don't have any beta program or anything similar to make it worth distinguishing between them, I guess :shrug: .

PR #481, opened 2020-04-16
docs: fix incomplete release process

izaera:
We were missing the tag because it was kind of hidden inside the next step, which we removed in the previous PR because it wasn't needed any more.

Note that I have added a command to push the exact tag directly to upstream. 

I tried in v9.5.1 release doing `git push upstream 9.x --follow-tags` to make everything in one step but only managed to:

1) Push a stale `generator-liferay-theme/vv10.0.0-alpha.1` that I had in local (I have fixed the disaster upstream; no need to worry).
2) Don't get the `v9.5.1` tag pushed upstream

:shrug: 

Thus, I decided that pushing the tag explicitly is safer for the future...

**Doubt: I'm unsure if pushing the tag will push the chore commit upstream too. If that was the case, the second push would be superfluous...** :point_left: @wincent Any idea?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,7 +70,7 @@ git grep 9.0.0-alpha.0
 
 Run `yarn changelog --version=v$VERSION`.
 
-### 4. Send a release PR
+### 4. Update release number and tag commit
 
 ```sh
 # See the tests pass locally:
@@ -79,6 +79,8 @@ yarn ci
 # Prepare and push final commit:
 git add -A
 git commit -m "chore: prepare $VERSION release"
+git tag v$VERSION -m v$VERSION
+git push upstream v$VERSION
 git push upstream 9.x
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,8 +80,7 @@ yarn ci
 git add -A
 git commit -m "chore: prepare $VERSION release"
 git tag v$VERSION -m v$VERSION
-git push upstream v$VERSION
-git push upstream 9.x
+git push upstream 9.x --follow-tags
 ```
 
 ### 5. Update the release notes


wincent:
Good catch!

> I'm unsure if pushing the tag will push the chore commit upstream too. If that was the case, the second push would be superfluous...

Without trying this, I believe it will push *the commit object* too. The `git-push` man page says:

> `tag <tag> means the same as refs/tags/<tag>:refs/tags/<tag>`

But nothing about it suggests that it will update the branch. ie. I expect the tag and the commit to arrive on the remote, but the branch will continue to point where it was previously.

So you still need the second push:

```
git tag v$VERSION -m v$VERSION
git push upstream v$VERSION
git push upstream 9.x
```

Having said that, I disagree with the proposed change overall. I still think `--follow-tags` is better. The only reason it pushed garbage for you was because you had garbage in your local repo, and that garbage happened to be a tag pointing at a commit that was reachable from the new `HEAD`. Now that you have cleaned up that garbage (with `git tag -d`), it won't happen again. If we can replace two steps with one, that reduces the possibility of user error.

So I'd go with:

```
git tag v$VERSION -m v$VERSION
git push upstream 9.x --follow-tags
```

Or, if you wanted to be paranoid:

```
git tag v$VERSION -m v$VERSION
git push upstream 9.x --follow-tags --dry-run
git push upstream 9.x --follow-tags
```

But at the end of the day, if a bad tag is committed, or tag isn't pushed, we can trivially fix it, so I don't think we need to be *that* careful.

izaera:
No, I didn't explain myself. I did a `follow-tags` after setting the tag and it only pushed the incorrect tag, but did nothing with the `v9.5.1` one :flushed: .

Then, to make sure, I removed the incorrect tag locally and upstream and tried again: `git push upstream 9.x --follow-tags` and nothing happened.

So I had to explicitly push the `v9.5.1` tag.

I agree that reading `--follow-tags` documentation it should be pushed, but it didn't. Don't know why.

But we can leave the `--follow-tags`. I think it should work and if not we'll see what happens next time.

izaera:
I will push the documentation with:

```
git tag v$VERSION -m v$VERSION
git push upstream 9.x --follow-tags
```

in some minutes.

wincent:
> Then, to make sure, I removed the incorrect tag locally and upstream and tried again: `git push upstream 9.x --follow-tags` and nothing happened.

It seems unlikely that you have found a bug in Git. Having used `--follow-tags` probably hundreds of times in the past, in my experience it has always done what it says it does in the man page:

       --follow-tags
           Push all the refs that would be pushed without this option, and also push
           annotated tags in refs/tags that are missing from the remote but are pointing at
           commit-ish that are reachable from the refs being pushed.

Hypothetically, if you were on the `master` branch when you pushed the first time that would explain why the first `--follow-tags` invocation would have pushed up the bad tag you had locally (because it was reachable from the `HEAD` of `master`). It would not push the `v9.5.1` tag because it is not reachable from the `HEAD` of `master`.

As for why `git push upstream 9.x --follow-tags` wouldn't work, that one is a total mystery to me. Even if you do a no-op `push` of something that is already up-to-date, `--follow-tags` will still push any tags that you have newly added and which are reachable. 🤔

izaera:
Never mind. We'll see on next release :shrug: 

I'll merge this.

wincent:
Ah, I know what the cause is. It's because you didn't make an annotated tag for v9.5.1. ie. you did `git tag v9.51` and left off the `-m v9.5.1`. Note the difference between these two:

## `git show v9.5.1`

```
commit 3221b794402f647c85d6d63f4d22265f35caf660 (tag: v9.5.1, upstream/9.x, upstream-rw/9.x)
Author: Iván Zaera Avellón <izaera@gmx.es>
Date:   Thu Apr 16 10:45:51 2020 +0200

    chore: prepare 9.5.1 release
```

## `git show v9.5.0`

```
tag v9.5.0
Tagger: Iván Zaera Avellón <izaera@gmx.es>
Date:   Thu Jan 30 17:34:49 2020 +0100

v9.5.0

commit 322efe3d55ad43dd417f441095ccb4f15dc1c884 (tag: v9.5.0)
Author: Iván Zaera Avellón <izaera@gmx.es>
Date:   Thu Jan 30 17:26:45 2020 +0100

    chore: prepare v9.5.0 release
```

`git push --follow-tags` won't push unannotated tags:

```
tmp.CmuvESzK ❯❯ git tag crap
tmp.CmuvESzK ❯❯ git ci --allow-empty -m 'a commit'
[master 12f45d3] a commit
tmp.CmuvESzK ❯❯ git push --follow-tags
Enumerating objects: 1, done.
Counting objects: 100% (1/1), done.
Writing objects: 100% (1/1), 185 bytes | 185.00 KiB/s, done.
Total 1 (delta 0), reused 0 (delta 0)
To github.com:wincent/test.git
   79e35b2..12f45d3  master -> master
tmp.CmuvESzK ❯❯ git tag wonderful -m annotation
tmp.CmuvESzK ❯❯ git push --follow-tags
Enumerating objects: 1, done.
Counting objects: 100% (1/1), done.
Writing objects: 100% (1/1), 168 bytes | 168.00 KiB/s, done.
Total 1 (delta 0), reused 0 (delta 0)
To github.com:wincent/test.git
 * [new tag]         wonderful -> wonderful
```

This is another reason to use `yarn version` to create releases, like we have in other repos, because it eliminates this kind of mistake. But as we discussed previously, v9 is versioned "in block", so using `yarn version` isn't going to work without jumping through some hoops.

👍 to merging

Issue #482, opened 2020-04-16
Inheritance of themes works reversed when more than one level exists


izaera:
See [originating issue](https://issues.liferay.com/browse/LPP-36970).

Basically, if you have three themes: `a`, `b` and `c` and (`c` extends `b` extends `a`), when you build `c` files from `a` take precedence over those from `b`.

izaera:
# From the originating issue:

**Details:**
The customer created three themes (a, b and c) using: yo liferay-theme and installed them as global npm modules.

Next, they used gulp extend to make b-theme inherit from a-theme and finally for c-theme to inherit from b-theme.

To differentiate between the themes they added custom background-colours in _custom.scss. They also made overrides to the portal_normal.ftl templates but only in a- and b-theme to prove the issue that the inheritance isn't going in the correct order.

All they did in portal_normal.ftl was to add an h1 specifying the theme's name in the content section.

After extending the themes and finally building c-theme they would expect it to take the template file from b-theme but instead it's grabbing the one coming from a-theme.

It's not included in the attached example but they've tried to throw in a fourth theme in the mix as well to inherit from the third theme. The result then is for it to grab whatever is overridden in the second theme (not the third as you'd want it to but not the first either as when we had three themes).

**Details about expected behavior:**
They are expecting c-theme to have blue background-color as they're overriding _custom.scss. If they hadn't done that, they'd expect it to take the color set in b-theme as that's the one it inherits from. When it comes to the templates, they're expecting the same behavior when overriding. If they have a custom portal_normal.ftl in a-theme but none in b-theme or c-theme they'd expect both of the latter themes to use a-theme's portal_normal.ftl but when they override it in b-theme as in the attached example they'd expect both b-theme and c-theme to use the template from b-theme. Instead for some reason, c-theme grabs it from a-theme.

PR #483, opened 2020-04-16
fix: #482 reverse themes inheritance order

izaera:
Inheritance of themes is being applied in reverse order. This PR fixes that.

### DIFF
--- a/packages/liferay-theme-tasks/theme/lib/getBaseThemeDependencies.js
+++ b/packages/liferay-theme-tasks/theme/lib/getBaseThemeDependencies.js
@@ -35,9 +35,11 @@ function getBaseThemeDependencies(
 			baseTheme.name
 		);
 
+		dependencies = getBaseThemeDependencies(baseThemePath, dependencies);
+
 		dependencies.push(path.resolve(baseThemePath, 'src', '**', '*'));
 
-		return getBaseThemeDependencies(baseThemePath, dependencies);
+		return dependencies;
 	} else if (baseTheme === 'styled' || baseTheme === 'admin') {
 		dependencies.splice(
 			1,


wincent:
Code looks fine but it is hard for me to validate this just by looking at it. Could you describe what you did to test? As long as you're confident in it, this looks good to me.

izaera:
Ah OK. So, it's basically what they describe in the [LPP](https://issues.liferay.com/browse/LPP-36970):

> **Details**:
>The customer created three themes (a, b and c) using: yo liferay-theme and installed them as global npm modules.

>Next, they used gulp extend to make b-theme inherit from a-theme and finally for c-theme to inherit from b-theme.

>To differentiate between the themes they added custom background-colours in _custom.scss. They also made overrides to the portal_normal.ftl templates but only in a- and b-theme to prove the issue that the inheritance isn't going in the correct order.

>All they did in portal_normal.ftl was to add an h1 specifying the theme's name in the content section.

>After extending the themes and finally building c-theme they would expect it to take the template file from b-theme but instead it's grabbing the one coming from a-theme.

>It's not included in the attached example but they've tried to throw in a fourth theme in the mix as well to inherit from the third theme. The result then is for it to grab whatever is overridden in the second theme (not the third as you'd want it to but not the first either as when we had three themes).

>**Details about expected behavior**:
>They are expecting c-theme to have blue background-color as they're overriding _custom.scss. If they hadn't done that, they'd expect it to take the color set in b-theme as that's the one it inherits from. When it comes to the templates, they're expecting the same behavior when overriding. If they have a custom portal_normal.ftl in a-theme but none in b-theme or c-theme they'd expect both of the latter themes to use a-theme's portal_normal.ftl but when they override it in b-theme as in the attached example they'd expect both b-theme and c-theme to use the template from b-theme. Instead for some reason, c-theme grabs it from a-theme.

izaera:
The problem with the code was that the changed function is recursive and collects globs like `<package dir>/src/**/*` to pass all the files of a theme project through gulp.

As it is written, the calls for `a`, `b` and `c` (if c extends from b extends from a) go like this:

getDeps(c) calls getDeps(b) call getDeps(a)

But, for example, when the call getDeps(b) invokes getDeps(a) it puts its own globs (b/src/**/*) before a's globs making a's files override those in b (which is exactly the opposite of what we want).

izaera:
The test was reproducing the scenario described in the originating issue (with a, b and c themes) and fix it.

wincent:
> The test was reproducing the scenario described in the originating issue (with a, b and c themes) and fix it.

That's what I wanted to hear. Go for it!

Issue #466, opened 2020-04-07
Backport --config option to v8


izaera:
This is backport of #455 to v8 and v9.

I'm not quite sure if v9 should be maintained or deprecated, given that v10 will target the same platforms... Any opinions? :thinking:

wincent:
v9 should not be maintained; v10 *replaces* it.

v8 should be maintained. It does not have a replacement.

izaera:
This was released as `v8.2.0` and has been tested by @david-truong, confirming that it works.

Closing.

PR #484, opened 2020-04-22
feat: #466 backport batch mode (--config) to v8

izaera:
I have cherry-picked the `promptWithConfig` method from v10 and adapted it to v8. It mostly resembles the same code in v10 but it's a bit different due to the refactors I did in v10.

I have tested the `--config` option manually and compared its output to that of the interactive generator. All seems right. 

However, because this will be tested by the blade team when using it, I have not been very exhaustive in my tests. The plan is releasing an alpha version and then, when the blade team gives the OK, release the final minor version upgrade.

### DIFF
--- a/packages/generator-liferay-theme/generators/app/index.js
+++ b/packages/generator-liferay-theme/generators/app/index.js
@@ -6,6 +6,7 @@
 
 'use strict';
 
+const fs = require('fs');
 const _ = require('lodash');
 const chalk = require('chalk');
 const Insight = require('insight');
@@ -15,6 +16,9 @@ const Generator = require('yeoman-generator');
 const yosay = require('yosay');
 
 const lookup = require('liferay-theme-tasks/lib/lookup');
+
+const config = require('../../lib/utils/config');
+const promptWithConfig = require('../../lib/utils/promptWithConfig');
 
 module.exports = class extends Generator {
 	initializing() {
@@ -49,7 +53,12 @@ module.exports = class extends Generator {
 		const insight = this._insight;
 
 		if (_.isUndefined(insight.optOut)) {
-			insight.askPermission(null, _.bind(this._prompt, this));
+			if (config.batchMode()) {
+				insight.optOut = true;
+				this._prompt();
+			} else {
+				insight.askPermission(null, _.bind(this._prompt, this));
+			}
 		} else {
 			this._prompt();
 		}
@@ -142,14 +151,11 @@ module.exports = class extends Generator {
 
 		if (!skipInstall) {
 			this.on('npmInstall:end', () => {
-				const gulp = require('gulp');
-
-				// TODO: remove in v9
-				// See: https://github.com/liferay/liferay-js-themes-toolkit/issues/196
-				process.argv = process.argv.slice(0, 2).concat(['init']);
-
-				require('liferay-theme-tasks').registerTasks({gulp});
-				gulp.start('init');
+				if (config.batchMode()) {
+					this._runBatchGulpInit();
+				} else {
+					this._runGulpInit();
+				}
 			});
 
 			this.installDependencies({bower: false});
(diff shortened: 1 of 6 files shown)


izaera:
The CI is failing because I'm using `{...object}` destructuring. 

I will change it to `Object.assign()` to be safer, given that this code is very old and may be executed with old node versions.

packages/generator-liferay-theme/lib/utils/config.js
```diff
@@ -1,17 +1,19 @@
 /**
- * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * © 2017 Liferay, Inc. <https://liferay.com>
+ *
```
wincent, line 3: These changes seem wrong; this is supposed to match [copyright.js](https://github.com/liferay/liferay-js-themes-toolkit/blob/master/copyright.js):

```
/**
 * SPDX-FileCopyrightText: © <%= YEAR %> Liferay, Inc. <https://liferay.com>
 * SPDX-License-Identifier: MIT
 */
```

wincent: Ah no, sorry. Forgot that `master` and `8.x` aren't in sync on that one.

packages/generator-liferay-theme/generators/app/index.js
```diff
@@ -338,6 +357,70 @@ module.exports = class extends Generator {
...
+			'liferay-theme.json',
+			JSON.stringify({LiferayTheme: answers}, null, 2)
+		);
+	}
+
+	_runGulpInit() {
+		const gulp = require('gulp');
+
+		// TODO: remove in v9
+		// See: https://github.com/liferay/liferay-js-themes-toolkit/issues/196
```
wincent, line 417: You can remove this comment, I think? Related issue is closed/fixed in v9.

packages/generator-liferay-theme/generators/layout/index.js
```diff
@@ -226,6 +253,70 @@ module.exports = class extends Base {
...
+			'liferay-plugin.json',
+			JSON.stringify({LiferayPlugin: answers}, null, 2)
+		);
+	}
+
+	_runGulpInit() {
+		const gulp = require('gulp');
+
+		// TODO: remove in v9
+		// See: https://github.com/liferay/liferay-js-themes-toolkit/issues/196
```
wincent, line 313: KIll.

packages/generator-liferay-theme/lib/utils/config.js
izaera: Oops. There must be something wrong in my environment....

packages/generator-liferay-theme/generators/app/index.js
izaera: I though the same but was unsure. I guess we can remove it.

packages/generator-liferay-theme/generators/layout/index.js
izaera: Bill.

izaera:
> wincent: lg2m modulo some minor nits. Hope you tested this thoroughly... crossed_fingers

> izaera (some lines up):  I have not been very exhaustive in my tests

:shrug:

izaera:
However, as I said:

>  this will be tested by the blade team when using it,

Those will be the final tests. Also, because of the nature of the change, unless someone uses `--config` the toolkit will keep working as always.

wincent:
> However, as I said:
> 
> > this will be tested by the blade team when using it,
> 
> Those will be the final tests. Also, because of the nature of the change, unless someone uses `--config` the toolkit will keep working as always.

Yeah, *that's* the bit I was worried about (unintentionally breaking the non-`--config` use case). If `--config` doesn't work it doesn't really matter because they will tell us and we will fix it, and nobody will be hurt.

Issue #485, opened 2020-04-22
Temporary compile-css files race condition


izaera:
Currently the build process creates a temporary `_bourbon.scss` file in `/tmp/tmp` which may lead to race conditions if several builds are run in parallel.

These files [were created inside `node_modules`](https://github.com/liferay/liferay-js-themes-toolkit/blob/fe8c7c5348fb47cb757a8f476fab0d784804545f/packages/liferay-theme-tasks/lib/bourbon_dependencies.js#L25) in previous versions of the project, what seems a quite dirty technique. 

That was the reason why it was changed introducing this bug exactly in [this modification](https://github.com/liferay/liferay-js-themes-toolkit/commit/bb63f0d023517c84202bf3c5efaa76f77c8e34a5#diff-fb29d65ae0a3361a552dfb08f9d72ff0R25).

izaera:
BTW: this was detected during a portal's CI build (see https://github.com/izaera/liferay-portal/pull/138)

wincent:
🤦‍♂️ — and the "dirty technique" [was in use for 5 years](https://github.com/liferay/liferay-js-themes-toolkit/commit/c8e77d6ab6e4f77505d5a48a5ef98d2e4509866d) 😂

izaera:
Yes :shrug:

PR #486, opened 2020-04-22
fix: #485 move temporary files out of /tmp

izaera:
See issue #485 for a description of the issue and the commit description for a description of the fix.

### DIFF
--- a/packages/liferay-theme-tasks/lib/project/index.js
+++ b/packages/liferay-theme-tasks/lib/project/index.js
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: MIT
  */
 
+const crypto = require('crypto');
 const fs = require('fs-extra');
 const _ = require('lodash');
 const path = require('path');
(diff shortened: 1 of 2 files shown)


packages/liferay-theme-tasks/lib/project/index.js
```diff
@@ -62,6 +63,28 @@ class Project {
...
+	 * Each call to this method will return a new directory.
+	 *
+	 * The returned directory is guaranteed to exist during the execution of the
+	 * current build but no more than that. However, it is not deleted on exit.
+	 *
+	 * @return {FilePath}
+	 */
+	tmpdir() {
+		const tmpdir = this.options.pathBuild.join(
+			'.tmp',
```
wincent, line 79: Why is this `.tmp` as opposed to, say, `tmp`?

izaera: To avoid mixing it with legitimate content that is put into the final war. That way it doesn't appear when you `ls` the folder and it is seems a bit clearer that it is a temporary non-final thing, I hope.

izaera: The solution I have applied to this problem in the JS Toolkit has been creating several folders inside `build` as gradle does in the portal build, but I didn't want to change the behaviour of the Themes Toolkit so much.

Do you think it would be better to do that here, too? :thinking:

wincent: > Do you think it would be better to do that here, too? 🤔

No, this is all "implementation detail" kind of stuff, so I don't think there's any great need or urgency to change it.

Your answer effectively boils down to: "because dot-files don't get compiled into JARs", which is reason enough for me. I just didn't know how it worked.  👍

Issue #445, opened 2020-03-13
Deploy Layout Template failed with error "Invalid output folder"


zhangyan0701:
**Steps to reproduce:**
1. Install the latest version of theme generator (v9.5.0).
2. Run `yo liferay-theme:layout` to generate a layout template of 7.3:
- Assert What would you like to call your theme?: My Liferay Test Layout
- Would you like to use this as a themeID? my-liferay-test-layout
- Which version of Liferay is this theme for? 7.3
- How many columns would you like for row 1? 1
- How wide do you want row 1 column 1? 12/12 - 100%
- Assert npm install is run
- Select your deployment strategy: Local App Server
- Enter the path to your app server directory: {insert tomcat directory here}
- Enter the url to your production or development site: http://localhost:8080
3. **Assert there is no log like "‘init’ after xx μs" in console.**
4. Navigate to the theme folder.
5. Run `gulp deploy`.

**Expected Result:**
Layout template is deployed to portal.

**Actual Result:**
Deploy failed with below error:
```
[17:52:44] Starting 'plugin:deploy'...
[17:52:44] 'plugin:deploy' errored after 687 μs
[17:52:44] Error: Invalid output folder
    at Gulp.dest (C:\Users\liferay\dandanlayout-layouttpl\node_modules\vinyl-fs\lib\dest\index.js:15:11)
    at Gulp.gulp.task (C:\Users\liferay\dandanlayout-layouttpl\node_modules\liferay-theme-tasks\plugin\tasks\deploy.js:28:15)
    at module.exports (C:\Users\liferay\dandanlayout-layouttpl\node_modules\orchestrator\lib\runTask.js:34:7)
    at Gulp.Orchestrator._runTask (C:\Users\liferay\dandanlayout-layouttpl\node_modules\orchestrator\index.js:273:3)
    at Gulp.Orchestrator._runStep (C:\Users\liferay\dandanlayout-layouttpl\node_modules\orchestrator\index.js:214:10)
    at Gulp.Orchestrator.start (C:\Users\liferay\dandanlayout-layouttpl\node_modules\orchestrator\index.js:134:8)
    at runNextSet (C:\Users\liferay\dandanlayout-layouttpl\node_modules\run-sequence\index.js:124:15)
    at Gulp.onTaskEnd (C:\Users\liferay\dandanlayout-layouttpl\node_modules\run-sequence\index.js:102:4)
    at Gulp.emit (events.js:198:15)
[...]

PR #488, opened 2020-05-07
fix: #445 gulp init failure

izaera:
`gulp init` for layout projects was running the `liferay-theme-tasks` associated to themes. This caused the deployment directory to be written in `liferay-theme.json` instead of `liferay-plugin.json` making `gulp deploy` fail later.

Of course, the fix has been tested. Following the steps described in the issue.


izaera:
@wincent I knew it, so I added it in a second edit. :joy:

Issue #493, opened 2020-06-25
Extending from a URL themelet doesn't work properly


izaera:
When using v8.0.0 of the Liferay Theme Generator the gulp extend command does not work as expected. Instead it seemingly inserts theme information a second time into the resulting package.json file.

This behavior does not occur in the latest version of the theme generator however the latest version does not support making a theme for DXP 7.1 which is the customer’s use case.

**Steps to reproduce:**

1. Install theme generator: `npm install -g generator-liferay-theme@^8.0.0`
2. Run `yo`,
select: Liferay Theme
name: test
id: test
version: 7.1
deployment: other
url: default

3. Call `gulp extend` from the root of the created theme
kind: themrlet
Where: url
url: https://registry.npmjs.org/sign-in-modal-themelet/-/sign-in-modal-themelet-1.0.2.tgz

**Results:**

Before running the gulp extend command the package.json looks like this:

```js
{
	...,
	"liferayTheme": {
		"baseTheme": "styled",
		"screenshot": "",
		"rubySass": false,
		"templateLanguage": "ftl",
		"version": "7.1",
		"hookModules": [
			"liferay-theme-es2015-hook"
		],
		"themeletDependencies": {}
	},
	"main": "package.json",
	"name": "fptc-t-child-theme",
	"version": "1.0.0",
	"dependencies": {}
}
```

After extension it looks like this:

```json
	...,
	"liferayTheme": {
		"baseTheme": {
			"liferayTheme": {
				"screenshot": "",
				"themelet": true,
				"version": "*"
			},
			"name": "sign-in-modal-themelet",
			"version": "1.0.2"
		},
		"screenshot": "",
		"rubySass": false,
		"templateLanguage": "ftl",
		"version": "7.1",
		"hookModules": [
			"liferay-theme-es2015-hook"
		],
		"themeletDependencies": {}
	},
	"main": "package.json",
	"name": "fptc-t-child-theme",
	"version": "1.0.0",
	"dependencies": {}
```

However, it should look like these:

```json
	...,
	"liferayTheme": {
		"baseTheme": "styled",
		"screenshot": "",
		"rubySass": false,
		"templateLanguage": "ftl",
		"version": "7.1",
		"hookModules": [
			"liferay-theme-es2015-hook"
		],
		"themeletDependencies": {
			"sign-in-modal-themelet": {
				"liferayTheme": {
					"screenshot": "",
					"themelet": true,
					"version": "*"
				},
[...]

PR #494, opened 2020-06-25
fix: #493 make extending from URL work

izaera:
I have copied the missing code from master following the same structure found in other prompts to make comparing code easier.

I have tested that it works manually (following the bug steps and checking the output).

### DIFF
--- a/packages/liferay-theme-tasks/lib/prompts/url_package_prompt.js
+++ b/packages/liferay-theme-tasks/lib/prompts/url_package_prompt.js
@@ -16,18 +16,9 @@ class URLPackagePrompt {
 
 	init(config, cb) {
 		this.done = cb;
+		this.themelet = config.themelet;
 
-		inquirer.prompt(
-			[
-				{
-					message: 'Enter the URL for the package:',
-					name: 'packageURL',
-					type: 'input',
-					validate: this._validatePackageURL,
-				},
-			],
-			this._afterPrompt.bind(this)
-		);
+		this._prompt();
 	}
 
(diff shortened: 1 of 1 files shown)


packages/liferay-theme-tasks/lib/prompts/url_package_prompt.js
```diff
@@ -44,14 +35,44 @@ class URLPackagePrompt {
...
-		answers.module = config.name;
+
 		answers.modules = {
 			[config.name]: Object.assign({}, config, {
 				__packageURL__: answers.packageURL,
 			}),
 		};
+
+		if (this.themelet) {
+			answers.module = {[config.name]: true};
```
wincent, line 46: My only concern with this would be the use of a computed property like this, as customers using older DXP, older themes toolkit, might also be using older node.

https://kangax.github.io/compat-table/es6/#object_literal_extensions_computed_properties doesn't say when support was added, but it does note that at least NodeJS v8 supports it.

Ah, [looks like support was added in v4](https://stackoverflow.com/questions/31498206/how-to-use-es6-computed-property-names-in-node-iojs#comment59071864_31507859), so we're good.

izaera: I tested with node 10 and didn't fail. But good to know that it works from 4 on :+1:

izaera: In fact, the same syntax is used above in currently existing code, so...

Issue #495, opened 2020-06-29
Support yarn workspaces in font-awesome enabled projects


izaera:
Currently `font-awesome` enabled themes fail when building whenever the `font-awesome` package is hoisted to the workspace's `node_modules` folder.

izaera:
We may need to rollback this because, even if it is correct, we may want to fix it in a different upper level. Otherwise we may end up adding code for every possible scenario we may think of, which doesn't seem reasonable.

izaera:
@dave-truong has confirmed that this fix works. However, we think it is not a good idea to put logic to support every possible scenario (project setup) in this level, i.e., we thing the Themes Toolkit should be agnostic on your project's setup and just do what it does best: bundling a theme.

So, what we'll do, unless anyone has anything against it, is providing an option to pass SASS include paths to the themes toolkit. Of course, I will check first if we have anything similar available.

jbalsas:
We already support passing paths and we use it in several places... it worries me that this is an ootb feature that won't work in yarn by default. To remove the fix we added, we'd need to guarantee that our generated code goes through a safe wrapper like npm-scripts or similar in order to ensure the proper path configuration.

izaera:
So, it's true that, until we can generate projects that are built with an upper tool layer controlling the Themes Toolkit (whether it is `blade`, `liferay-npm-scripts`, or anything similar...) we cannot avoid failures for projects which don't pass the `sass-include-paths` argument to the Themes Toolkit.

So, until we get to the state where the Themes Toolkit is never run alone, but in a managed execution, we have to leave this fix as a workaround.

PR #496, opened 2020-06-29
fix: #495 support yarn workspaces

izaera:
Support yarn workpaces in font-awesome enabled projects by simply adding the
workspace's node_modules folder to the SASS include paths.


izaera:
Going to merge this as I need to make a release and there's no one available to review.

PR #518, opened 2020-09-28
fix: download 4.x series of admin/classic themes

izaera:
This is necessary since there was a change in the portal's source code that switched version numbers from 3.x to 4.x.

This PR also adds a feature consisting of a `--themeVersion` argument that can be given in the CLI to force the generator to download a specific version instead of choosing the latest of the target series. This is useful in scenarios where the build must be repeatable or where maven repository metadata must be overriden for any reason.

This can be tested easily by running:

`yo /path/to/your/themes/toolkit/project/packages/generator-liferay-theme/generators/classic/index.js`

and:

`yo /path/to/your/themes/toolkit/project/packages/generator-liferay-theme/generators/classic/index.js --themeVersion 4.0.0`


packages/generator-liferay-theme/lib/generation/theme-based.js
```diff
@@ -30,25 +31,29 @@ async function writing(generator, themeName) {
...
-		'',
-		info`
-		Querying https://mvnrepository.com for the latest version of Liferay's
-		${themeName} theme compatible with ${liferayVersion}...
-		`
-	);
+	let {themeVersion} = argv;
 
-	const themeVersions = await getThemeVersions(themeName);
+	if (!themeVersion) {
```
jbalsas, line 36: I'm just a bit confused as to why we do `!themeVersion` here but `themeVersion === undefined` down below...

Other than that, LGTM!

izaera: :+1:

PR #524, opened 2020-10-15
fix: [TS] LPS-121779 race condition on liferay-theme-tasks gulp watch task

georgel-pop-lr:
LPS-121779 Fix race condition on liferay-theme-tasks gulp watch task 'Cannot read property 'path' of undefined' when changes saved in short time, multiple times.

### DIFF
--- a/packages/liferay-theme-tasks/theme/tasks/watch.js
+++ b/packages/liferay-theme-tasks/theme/tasks/watch.js
@@ -305,8 +305,6 @@ module.exports = function() {
 
 			const taskArray = getBuildTaskArray(resourceDir);
 
-			taskArray.push(clearChangedFile);
-
 			runSequence(...taskArray);
 		});
 	}


izaera:
@georgel-pop-lr This PR is failing the semantic PR CI test. Can you please check our [Pull Request Etiquette doc](https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/pull_requests.md) and fix it accordingly? Thx :-)

georgel-pop-lr:
> @georgel-pop-lr This PR is failing the semantic PR CI test. Can you please check our [Pull Request Etiquette doc](https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/pull_requests.md) and fix it accordingly? Thx :-)

Done, thanks @izaera

izaera:
Tests are green and I assume @georgel-pop-lr has tested this, so LGTM.

I've been investigating the removed line and it looks like it was brought from the `8.x` branch by [this commit](https://github.com/liferay/liferay-js-themes-toolkit/commit/c15b0ed1a20fe9810818917f110bff1f84ae7ac9). That was a big refactor to make watch work correctly (because it had a lot of synchronization issues before).

However, I'm unsure of why removing that line fixed the problem. Can you please elaborate a bit more @georgel-pop-lr so that we know in the future why that was removed? The best place to do it would be the commit message IMO, so that it can be examined without connecting to the web with a simple git blame or log.

Would you mind adding that an pushing force the PR again so that we merge it after?

Also, I think this could be merged in maintenance branches too... What do you think @wincent ? Should we do the backport now or on demand?

Doing it now implies more work now for something that may not fail in the future, but on the other hand, if we don't backport now and it fails in the future, we will have to spend more time.

Alternatives that come to my mind are:

1. Backport now
2. Lazy backport
3. Maybe better: tag backportable issues somehow so that they are available and are the first  thing we check when we see an issue in older versions

wincent:
> Doing it now implies more work now for something that may not fail in the future, but on the other hand, if we don't backport now and it fails in the future, we will have to spend more time.

I think that's the key right there: for each change, you have to make a prediction for whether it will fail in the future — if you think the problem affects other branches, then the right thing to do is fix all the branches; if you don't think the problem affects other branches, then you don't touch the other branches.

Obviously with both options there is the possibility of a false negative or a false positive:

- You think it affects other branches but it doesn't, so you wasted your time porting it and potentially even risked a regression just because you touched it.
- You think it doesn't affect other branches but it does, so you eventually find that out and end up doing the same fix again later.

Either way, I given that we don't have a crystal ball, I don't think either of these possibilities really influence the decision.

georgel-pop-lr:
> Tests are green and I assume @georgel-pop-lr has tested this, so LGTM.
> 
> I've been investigating the removed line and it looks like it was brought from the `8.x` branch by [this commit](https://github.com/liferay/liferay-js-themes-toolkit/commit/c15b0ed1a20fe9810818917f110bff1f84ae7ac9). That was a big refactor to make watch work correctly (because it had a lot of synchronization issues before).
> 
> However, I'm unsure of why removing that line fixed the problem. Can you please elaborate a bit more @georgel-pop-lr so that we know in the future why that was removed? The best place to do it would be the commit message IMO, so that it can be examined without connecting to the web with a simple git blame or log.
> 
> Would you mind adding that an pushing force the PR again so that we merge it after?
> 
> Also, I think this could be merged in maintenance branches too... What do you think @wincent ? Should we do the backport now or on demand?
> 
> Doing it now implies more work now for something that may not fail in the future, but on the other hand, if we don't backport now and it fails in the future, we will have to spend more time.
> 
> Alternatives that come to my mind are:
> 
> 1. Backport now
> 2. Lazy backport
> 3. Maybe better: tag backportable issues somehow so that they are available and are the first  thing we check when we see an issue in older versions

Hi @izaera,

Thanks for looking into this, the solution was both tested on my local and confirmed by our client, and works, not sure if can create any regression. 
Basically I saw adding logs, that we call operation in a not synchronize way, we do multiple save operation, in a short time, the task reload is being called and puts in state the changed file name, because we don't sync our operations,  'clearChangedFile' from a previous reload is being called and removes the actual state, when the actual reload is happening there is no changedFile available so the race condition is happening. 
About the commit message I thought is pretty explanatory, but if you don't think, so I will try to change it in a more proper way. I accept suggestions :D 
Regarding affecting other branches, well the client is in a previous version "liferay-theme-tasks": "8.2.2" so was affected, but depends on how they are using this library, I think they had automatic saving in their settings (they did not confirm this), but I was able to reproduced manually pretty easy.

Regards.

izaera:
@georgel-pop-lr I was referring to writing the explanation above in the body part of the commit itself :-). Don't worry about long commit messages, there's a subject and a body in git commit messages to account for that, so it's better to err on making the message too long than making it too small.

I'm going to merge this given that you have added the description to the PR and that's enough (because there's traceability to the PR). However, I prefer to have key information both in PRs and commits (or just in the commits, if you don't want to duplicate). I don't know about the others...

Thks for everything, @georgel-pop-lr :+1:

georgel-pop-lr:
> @georgel-pop-lr I was referring to writing the explanation above in the body part of the commit itself :-). Don't worry about long commit messages, there's a subject and a body in git commit messages to account for that, so it's better to err on making the message too long than making it too small.
> 
> I'm going to merge this given that you have added the description to the PR and that's enough (because there's traceability to the PR). However, I prefer to have key information both in PRs and commits (or just in the commits, if you don't want to duplicate). I don't know about the others...
> 
> Thks for everything, @georgel-pop-lr

Thanks @izaera I will take this into account into feature commits, Regards

wincent:
> However, I prefer to have key information both in PRs and commits (or just in the commits, if you don't want to duplicate). I don't know about the others...

I share your preference. Having it in the commit message is super valuable because you can search the history with `git log --grep`. I usually write the info in Markdown syntax (or close enough to it, so that it looks good when wrapped to 80 columns on the terminal), and then I copy-paste it into the PR (if necessary, applying some small edits to make it look good in the GitHub UI).

PR #534, opened 2020-12-03
chore(8.x): remove liferay-theme-mixins package

wincent:
We have three identical copies of this (differing only in version number) on three branches:

- `8.x` branch: this branch, last released as v8.2.2.
- `9.x` branch: the 9.x series, last released as v9.5.2.
- `master` (ie. 10.x series): the 10.x series, never released and still labeled as v9.5.0.

Now, we have no internal references to this package in this repo, no recent or extensive usage of it in liferay-portal (most recent usage can be found in the `osb-faro` directory of the 7.1.x branch), and a Google search turns up no obvious written documentation or encouragement to use it. Finally, the package itself is basically just 4 Sass files that haven't changed at all in recent years (at least, no changes between current `HEAD` of `master` and the oldest tag in the repo, `v8.0.0`, now two years old).

The last remaining reference is as a transitive dependency in this branch. As `yarn why` shows:

    => Found "liferay-theme-mixins@6.2.3"
    info Reasons this module exists
       - "_project_#liferay-theme-tasks#liferay-theme-deps-6.2" depends on it

There is no tag for v6.2.3 (which was released 5 years ago) in this repo, but we can still perform a rough comparison:

    mkdir old new
    (cd old && yarn init -y && yarn add liferay-theme-mixins@6.2.3)
    (cd new && yarn init -y && yarn add liferay-theme-mixins@9.5.2)
    git diff -w old new

As [the diff shows](https://gist.github.com/wincent/6593d563c4b32f1c9d5e82345b9414c4), the newer versions have some additional mixins and variables defined (as well as moving some stuff around) but there are no changes to the one prexisting mixin (`@respond-to`). In other words, additive changes only, and it seems exceedingly unlikely we'll ever need to release this one again.

All of which is to say that this is an opportunity for us to do some clean up.

1. (This commit) Remove package from `8.x` branch.
2. (Future PR targeting `master` branch) Remove package (seeing as `master` version is older than `9.x` version).
3. Migrate repo into liferay-frontend-projects monorepo.
4. Remove package from `master` branch `maintenance` directory (ie. where the 9.x series will reside); we do this as a final step to reduce clutter, but only after ensuring it's reflected in the history (so that searches for `liferay-theme-mixins` will return a hit which links back to this explanation).

Actually, now that I write that out I realize we can actually swap the order of 3/4 just fine because it will still show up in the history either way and will make the migration easier.

More context in [Slack discussion](https://liferay.slack.com/archives/G3JBR21HA/p1606995401279400).


wincent:
9.x change: https://github.com/liferay/liferay-js-themes-toolkit/pull/535

wincent:
10.x change: https://github.com/liferay/liferay-js-themes-toolkit/pull/536

wincent:
Travis is totally borked right now, but local `ci` is good, so going to merge this:

Screenshot 2020-12-03 -151204-zj5usMe2@2x

PR #536, opened 2020-12-03
chore(10.x): remove liferay-theme-mixins package

wincent:
This is the 10.x equivalent of [the corresponding 8.x](https://github.com/liferay/liferay-js-themes-toolkit/pull/534) and [9.x](https://github.com/liferay/liferay-js-themes-toolkit/pull/535) changes.

As described in detail there:

- We have three identical copies (differing only in version number) across three branches.
- We don't directly use or depend on the package in any branch of this repo, or in liferay-portal. (Most recent use was in `osb-faro` on the DXP 7.1.x branch, where it is never going to be touched.)
- The package hasn't changed in any salient way since at least v8.0.0 (two years ago).
- The 8.x series depends, transitively, on an older version of the package (v6.2.3, from five years ago). The dependency that brings this old version in is liferay-theme-deps-6.2, which we can safely say we are never going to touch again.

In order to ease migration of this repo into the liferay-frontend-projects monorepo, we're cleaning up the package. It's there in the history if we ever need to resurrect it, but we never will...


wincent:
Travis is backed up, but local `ci` is green, so going to merge this:

Screenshot 2020-12-03 -151551-LDvILbSp@2x

